PR #2, opened 2023-07-05
Create clouds.yaml and set proper variables

arxcruz:
This patch changes the following:
* Add clouds.yaml as template, passing the KeystoneApiEndpoint and OpenStackAdminPassword that is stored in the secrets. We are not using the keystone-operator to get the api endpoint yet, because it requires webhooks that are not yet implemented.
* Update the job to not restart on failure, otherwise it will execute tempest all the time, and that's not what we want.
* Update tempest-secret.yaml to add the KeystoneApiEndpoint variable.
* Update the /var/lib/config-data container to be read and write for the user, this will allow the tempest container to copy the clouds.yaml to the right directory.


controllers/tempest_controller.go
```diff
@@ -424,6 +425,8 @@ func (r *TempestReconciler) generateServiceConfigMaps(
 	templateParameters["ImageRefAlt"] = "image_alt_uuid"
 	templateParameters["IdentityUriV3"] = "http://keystone-public-openstack.apps-crc.testing"
 	templateParameters["AdminPassword"] = "adminpass"
```
slawqo, line 427: do we still need AdminPassword and IdentityUriV3 now when we have what You have added below?

arxcruz: I left that for now, I'm thinking in the future to add an option if the person want's to use the built-in tempestconf from the tempest container, or if they want to skip it and pass all the info manually. What do you think?

slawqo: it was there just as a first step I think but we should rely on the info provided in the CR's yaml file I think. Maybe we can add some default values for it in the API definition but I don't think we should hard code it here

arxcruz: Ok, I'll remve the whole config and the tempest.conf since it's not being used.

PR #3, opened 2023-07-07
Add support to Skiplist and Allowed list and Kolla

arxcruz:
This patch adds the following:
* Add the possibility to pass a list of tests to be skipped
* Add the possibility to pass a list of tests to be executed
* Update the Job to call kolla instead of directly use the run_tempest.sh from the container, this allow us to avoid make changes in the upstream container since it is also being used by other projects.
* Calling kolla we can use the tempest-config.json and copy the files to the right place and execute the run_tempest.sh, making it transparent for the user.
* Add include.txt and exclude.txt as template to list the tests, and set default to run tempest.api if none is passed (required otherwise the empty include.txt will make run the full set of tests)
* Update tempest-config.json to copy the files to the right places.


api/v1beta1/tempest_types.go
```diff
@@ -61,6 +61,18 @@ type TempestSpec struct {
 	// TempestRegex
 	TempestRegex string `json:"tempestRegex,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// AllowedTests
+    AllowedTests []string `json:"allowedTests,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// SkippedTests
+    SkippedTests []string `json:"skippedTests,omitempty"`
```
slawqo, line 70: please fix indentations for those 2 lines

slawqo (approved):

PR #4, opened 2023-09-12
kolla no longer respecting KOLLA_CONFIG_FILE

afazekas:
KOLLA_CONFIG_FILE is not respected by kolla [1].

Not sure how long we want to depend on kolla,
but it should be fixed for now.

[1] https://opendev.org/openstack/kolla/commit/5b1da017988c987fc68b55d1f45b5d2676474ce1


slawqo (approved):

afazekas:
recheck

PR #5, opened 2023-09-13
Using exesting secret for tempest

afazekas:
Using the same secret configmap which is generated by
the openstack-operator and also used by the openstackclient pods.


arxcruz (approved):

afazekas:
incomplete upload

templates/tempest/config/tempest-config.json
```diff
@@ -1,12 +1,6 @@
 {
     "command": "/var/lib/tempest/run_tempest.sh",
     "config_files": [
```
afazekas, line 3: This file renamed in pr4 , it might need to be re based.

kopecmartin:
tested, works
/lgtm

slawqo (approved):

PR #6, opened 2023-09-19
decouple container changes from operator changes

afazekas:
Using the tempest container just a library provider
besides providing the runtime dependencies only expecting a
service user 42480:42480 with a writable home directory.

Getting closer to openshift security guidelines
by not using root user.

This change allows us to either incubate a new main script
without having constant cross repo changes,
or to turn into templated main script usage.

The invoke_tempest is based on the script from the container,
just little bit simplified.

Not depending on kolla to copy files around.


arxcruz (approved):
lgtm

PR #7, opened 2023-09-20
Remove TempestRegexp argument

afazekas:
Looks like it was not in use since the begning.
Users usually finds simpler to have explicit negative list,
even tough it is possible to consutuct a singe regexp.
Also possible to construct a single regexp on the operator side,
in case other way of passing would be prefered later on.

Moving the default allowedTests to the defaults.

Note: Consider moving the tempest run arguments
to sub structure.


afazekas:
Added a second commit,
question do we want to structure the args deeper ?

kopecmartin:
lgtm, tested and it all still works .. we can add --regex and --smoke support in a new commit

kopecmartin:
it seems there is no bot? meh, i'll click the merge pull request button :D

PR #8, opened 2023-09-21
Init documentation

kopecmartin:
Prepare the documentation structure and configs required to build
a documentation.


afazekas:
Other projects using https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/.github/workflows/docs.yaml .

kopecmartin:
yep, seems like they publish the doc using GitHub Pages, another example:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/.github/workflows/documentation.yaml

Ok then, I'll propose a new patch set (or whatever it's called here) next week when I'm back from my PTO.

kopecmartin:
recheck

PR #10, opened 2023-10-25
Add tempest/tempestconf options

lpiwowar:
This patch introduces:

-  tempestconfRun section for the Tempest CR. This field allows a user to specify parameters with which discover-tempest-config should be executed inside the Tempest pod.

- a set of parameters for the tempestRun section in the Tempest CR. These new parameters allow a user to specify aruments with which tempest command should be executed inside the Tempest pod.

The documentation is updated as well so that it demonstrates how the new fields should be utilized.


lpiwowar:
Related patch for tcib: https://github.com/openstack-k8s-operators/tcib/pull/95

arxcruz (approved):
I am assuming we will change the container image to handle all these environments, or change the tempestconf (if it doesn't support these environments) otherwise, lgtm

lpiwowar:
@arxcruz here is the related PR for tcib => https://github.com/openstack-k8s-operators/tcib/pull/95

api/bases/test.openstack.org_tempests.yaml
```diff
@@ -133,6 +133,113 @@ spec:
...
+                    default: ""
+                    description: Generate Tempest accounts file.
+                    type: string
+                  debug:
+                    default: false
+                    description: Print debugging information.
+                    type: boolean
+                  deployerInput:
+                    default: ""
+                    description: Path to deployer file
```
afazekas, line 174: How you are going to map the deployer input into the container ?
Should we have predefined location and in case it present use it ?

afazekas (commented):
In case you are not using the generated entry point the change should remove it.

api/v1beta1/tempest_types.go
```diff
@@ -38,22 +38,162 @@ type Hash struct {
...
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=true
+	// Run tests in parallel
+	Parallel bool `json:"parallel,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=false
+	// Serial run
+	Serial bool `json:"serial,omitempty"`
```
slawqo, line 68: aren't serial and Parallel mutually exclusive? Can it be that both will be set to "True" for example?

lpiwowar: It is a good point. I was thinking about the same. But I'm not sure how much should the test-operator contain the logic of using the parameters. For me it is fine that the pod fails when the user decides to use the parameters incorrectly. But I guess it is up for a discussion (@kopecmartin, @afazekas) (?)

api/bases/test.openstack.org_tempests.yaml
lpiwowar: I think we've already discussed this. I'm adding it here for reference. The deployerInput is mapped into the tempest pod as a file using ConfigMap - [here](https://github.com/openstack-k8s-operators/test-operator/pull/10/commits/e9148516cda9fd11e8a180e5a799c2dbedc9ea42#diff-19fb6c8a471eb8f571e5e959d274704a582df8fa55a60f500593fcba5efb8823R421). The location of the file is set by the test-operator as well one line below.

api/v1beta1/tempest_types.go
kopecmartin: if a user passes both, --serial and --parallel, arguments, tempest will throw an error:
`tempest run: error: argument --parallel: not allowed with argument --serial/-t`

The way I see it, we don't need to check any argument compatibility on the operator/tempest image side - that's done by tempest itself. The operator together with the image definitions are responsible for getting the arguments from user to tempest - they are just a messenger.

docs/source/samples/tempestconf-config.yaml
kopecmartin, line 1: is there any specific reason we would want tempestconf configuration separated from tempest one? To me, it's one container thus it kind of makes sense it would be one config file. What happens when the operator will be able to run several containers (probably not at the same time) - tempest, tobiko, rally etc? We will end up with too many config files.

lpiwowar: Not really. To me, it seemed more readable this way but I can see your point. Do you want to update the doc so that we have a single config map?

lpiwowar: @kopecmartin the documentation is updated:)

kopecmartin:
tcib PRs got merged, we can proceed with this one now

PR #11, opened 2023-10-30
Allow to pass a PV as storage

afazekas:
This change allows to use an user managed pv to
use as results store.
We might use different strategies at the end.


templates/tempest/bin/invoke_tempest
```diff
@@ -4,7 +4,7 @@ set -x
 TEMPEST_DIR=$HOME/openshift
 OPERATOR_ETC=/etc/test_operator
 # NOTE: might be backed with a PV
-OUTPUT_PATH=$HOME/output
+OUTPUT_PATH=${TEMPEST_OUTPUTDIR:-$HOME/output}
```
lpiwowar, line 7: @afazekas do we want to still keep the invoke_tempest script? I think we should stick with only one script for tempest execution. And shouldn't we make the same change in the [run_tempest.sh](https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/tcib/base/os/tempest/run_tempest.sh) in tcib?

api/v1beta1/tempest_types.go
```diff
@@ -62,6 +63,15 @@ type TempestSpec struct {
 	// Tempest Container Image URL (will be set to environmental default if empty)
 	ContainerImage string `json:"containerImage"`
 
+	// +kubebuilder:validation:Optional
+	// persistentVolumeClaimVolumeSource represents a reference to a
+	// PersistentVolumeClaim in the same namespace. Usable for storing test results.
+        PersistentVolumeClaim *corev1.PersistentVolumeClaimVolumeSource `json:"persistentVolumeClaim,omitempty"`
```
lpiwowar, line 69: This is just an idea. I know that we might be changing this later but don't we want to choose more descriptive name? Something like testResultsPV?

templates/tempest/bin/invoke_tempest
lpiwowar: I'm removing the calling of the invoke_tempest script in [this](https://github.com/openstack-k8s-operators/test-operator/pull/10) PR - maybe I should update the description of the PR. I think we need to agree how we will approach this.

lpiwowar:
A similar issue was already resolved here -> https://github.com/openstack-k8s-operators/test-operator/pull/18. I'm closing this PR for now. If you think I'm wrong then please reopen this PR:).

PR #12, opened 2023-11-22
Add support for external plugins

lpiwowar:
This patch ensures that the test operator takes advantage of the new parameters of the tempest container:
        - TEMPEST_EXTERNAL_PLUGIN_GIT_URL
        - TEMPEST_EXTERNAL_PLUGIN_CHANGE_URL
        - TEMPEST_EXTERNAL_PLUGIN_REFSPEC
    
The user can use this parameters by specifying the following values for the Tempest CR:
        - externalPluginGitURL
        - externalPluginChangeURL
        - externalPluginRefspec


config/samples/test_v1beta1_tempest.yaml
```diff
@@ -23,6 +23,15 @@ spec:
     # smoke: false
     # serial: false
     # parallel: true
+    # externalPluginURL:
```
kopecmartin, line 26: so the first item in externalPluginRefspec is used together with the first one in externalPluginChangeURL and applied on first item from externalPluginURL?

What if I want to download barbican-tempest-plugin from master and a specific change from neutron-tempest-plugin? Do I need to write something like this?

```
externalPluginURL:
  - "https://opendev.org/openstack/barbican-tempest-plugin.git"
  - "https://opendev.org/openstack/neutron-tempest-plugin.git"
externalPluginChangeURL:
  - ""
  - "https://review.opendev.org/openstack/neutron-tempest-plugin"
externalPluginRefspec:
  - ""
  - "refs/changes/97/896397/2"
```

lpiwowar: _So the first item in externalPluginRefspec is used together with the first one in externalPluginChangeURL and applied on the first item from externalPluginURL?_

- Yes, exactly.

_What if I want to download barbican-tempest-plugin from master and a specific change from neutron-tempest-plugin?_

- If you want to apply changes only to a subset of repositories specified in externalPluginGitURL then you can do it like this:

```
externalPluginURL:
  - "https://opendev.org/openstack/neutron-tempest-plugin.git"
  - "https://opendev.org/openstack/barbican-tempest-plugin.git" # <- Put the repository without change at the end
externalPluginChangeURL:
  - "https://review.opendev.org/openstack/neutron-tempest-plugin"
externalPluginRefspec:
  - "refs/changes/97/896397/2"
```

OR like this:

```
externalPluginURL:
  - "https://opendev.org/openstack/neutron-tempest-plugin.git"
  - "https://opendev.org/openstack/barbican-tempest-plugin.git"
externalPluginChangeURL:
  - "-" # Use dash to indicate no change is needed
  - "https://review.opendev.org/openstack/neutron-tempest-plugin"
externalPluginRefspec:
  - "-" # Use dash to indicate no change is needed
  - "refs/changes/97/896397/2"
```

I updated the documentation so it is more clear. Maybe there is a better way to name the parameters (?) but nothing crosses my mind right now.

kopecmartin: I personally don't like the approach where we would have "-" - I'm almost 100% sure that users will not pass their plugins correctly :/ 

what about something like this?
```
externalPluginURL:
  - "https://opendev.org/openstack/barbican-tempest-plugin.git"
  - "https://opendev.org/openstack/neutron-tempest-plugin.git":
         - "https://review.opendev.org/openstack/neutron-tempest-plugin"
         - "refs/changes/97/896397/2"  
```

lpiwowar: Ok, I can modify it to accept this format. Shouldn't be that difficult. I will update the patch.

lpiwowar: @kopecmartin, so it turns out to be difficult after all :D. When you define externalPluginURL like this then there is no nice way how to define the type of externalPluginURL as it can be:

- a list of strings and 
- a list of dictionaries or 
- a combination of both. 

It can be either a list of strings or a list of dictionaries.  It would be probably possible to bypass this but I guess it would not be really openshifty. What do you think about this?

```
externalPlugin:
  - repository: "https://opendev.org/openstack/barbican-tempest-plugin.git"
  - repository: "https://opendev.org/openstack/neutron-tempest-plugin.git"
    changeRepository: "https://review.opendev.org/openstack/neutron-tempest-plugin"
    changeRefspec: "refs/changes/97/896397/2"
```

kopecmartin: this looks much better, i definitely like it more than the previous solution

kopecmartin (approved):
lgtm

PR #14, opened 2023-11-23
Update documentation

lpiwowar:
This is a quick update for the test-operator documentation. The following sections were udpated:

- Test Images
- User Guide
  - Running Operator Locally
  - Running Operator From Pre-Build Image
  - Custom Tempest Configuration

This patch also updates the workflow document that defines the docs job. The change ensures that the documentation jobs are executed only when a change to the documentation was made.

### DIFF
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -3,27 +3,113 @@ User Guide
 
 This guide is a starting point for configuring and running the `test-operator`.
 
-Installation
-------------
+
+Running Operator Locally Outside The Cluster
+--------------------------------------------
+This is **quick and easy way** how to experiment with the operator during development of a
+new feature.
 
 .. code-block:: bash
 
-    make install
+    make install run
 
-Execution
----------
-Execute the following command to run the operator. Note, that after running the
-following command you will need to switch to another terminal unless you run it
-in the background.
+Note, that after running the following command you will need to switch to
+another terminal unless you run it in the background.
 
-.. code-block:: bash
+Running Operator Using The Operator Lifecycle Manager (OLM)
+-----------------------------------------------------------
 
-    make run
+Another option is to start the operator by running the pre-build operator image
+stored on
+`quay.io <https://quay.io/repository/openstack-k8s-operators/test-operator>`_
+using the OLM.
 
 .. note::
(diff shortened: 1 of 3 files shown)


docs/source/guide.rst
```diff
@@ -3,27 +3,113 @@ User Guide
 
 This guide is a starting point for configuring and running the `test-operator`.
 
-Installation
-------------
+
+Running Operator Locally
+------------------------
+This is **quick and easy way** how to experiment with the operator during development of a
```
kopecmartin, line 9: This can be used even if a user doesn't want to experiment with the operator and they just wanna use it.

docs/source/guide.rst
```diff
@@ -48,8 +134,40 @@ To see the console output of the execution run the following:
...
+   admin_username = admin
+   admin_password = 1234
+   ...
+
 
-Getting logs
+Getting Logs
 ------------
-This is TBA.
+This is TB
```
kopecmartin, line 173: what does TB stand for?

docs/source/guide.rst
```diff
@@ -48,8 +134,40 @@ To see the console output of the execution run the following:
...
+
+    ---
+    apiVersion: test.openstack.org/v1beta1
+    kind: Tempest
+    metadata:
+        name: tempest-tests
+        namespace: openstack
+    spec:
+        containerImage: quay.io/podified-antelope-centos9/openstack-tempest:current-podified
+        tempestRun:
```
kopecmartin, line 153: this looks like something is missing there

docs/source/guide.rst
```diff
@@ -3,27 +3,113 @@ User Guide
...
+
+Running Operator Locally
+------------------------
+This is **quick and easy way** how to experiment with the operator during development of a
+new feature.
 
 .. code-block:: bash
 
-    make install
+    $ make install run
```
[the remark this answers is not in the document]
lpiwowar: Yes, I agree with you on this. I will remove it

docs/source/guide.rst
lpiwowar: To be :D ... I will fix it:)

PR #15, opened 2023-12-07
Mount TLS certificates to tempest container

lpiwowar:
This patch ensures we use tls certificate stored in secret named combined-ca-bundle when it is present in the openstack project.

When such a secret exists then we mount it to:
	- /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem.


kopecmartin (approved):
looks good to me

PR #18, opened 2024-01-09
Add test-operator-logs pvc

lpiwowar:
This patch ensures that test-operator creates a pvc called test-operator-logs and mounts it to pods spawned by the test-operator.

The pvc can be used as a persistent storage for all the artifacts related to the tests execution.


arxcruz (approved):
That is great news! That's the final piece we needed :) 
Great work Lukas!!!

kopecmartin (approved):
great job! Thanks Lukas

lpiwowar:
thank you, @arxcruz and @kopecmartin!

PR #19, opened 2024-01-10
Update RBAC rules for the test-operator

lpiwowar:
This patch updates the RBAC rules for the test-operator so that the controller can list persistent volume claims in the environment.

The change is needed in order to be able to sucessfully create pvcs from withing the operator.


kopecmartin (approved):
lgtm

PR #21, opened 2024-01-11
[docs] Fix name of CatalogSource

lpiwowar:
This patch fixes small typo in the documentation. We need to keep the name of the CatalogSource consistent. In the part of the documentation that is concerned with installation of the operator we use test-operator-index name and in the part of the documentation that describes the uninstallation of the operator we use test-operator-catalog.

This patch ensures that we use test-operator-catalog name for the catalogsource consistently.


kopecmartin (approved):
lgtm

PR #22, opened 2024-01-16
[docs] Update the Getting Logs section

lpiwowar:
The processing of the logs was updated in the test-operator. This patch updates the documenation which is concerned with accessing the logs produced during the test execution.


kopecmartin (approved):
thanks!

PR #24, opened 2024-01-16
Polish the doc about images used by the operator

kopecmartin:
Provide links to the tempest spec file as well as TCIB definition
for up to date list of dependencies and tempest plugins.


lpiwowar (approved):
Looks good to me!:)

PR #26, opened 2024-01-22
Refactor tempest code

lpiwowar:
This patch refactors the tempest code and prepares the code for other plugins like tobiko. The patch:

- Moves all code from the controller that can potentially be consumed by other plugins to common.go

- Removes files in templates folder as the way by which the spawned pods are configured is now different (we use env vars)

- Remove lines of code from the tempest_controller that are responsible for NetworkAttachments creation. These lines probably belong to the neutron-operator.


lpiwowar:
@kopecmartin This patch needs to be merged first as [this](https://github.com/openstack-k8s-operators/test-operator/pull/23) patch depends on it.

kopecmartin (approved):
looks good to me, loving the progress!

PR #27, opened 2024-01-22
[docs] Fix image source for logs pod

lpiwowar:
This is a small typo. The image for the logs pod contained valued "openstack" instead of "quay.io/quay/busybox".

### DIFF
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -277,14 +277,13 @@ For example this definition of Tempest object:
 will ensure that tempest will be executed with tempest.conf that looks like this:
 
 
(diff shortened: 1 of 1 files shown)


docs/source/guide.rst
```diff
@@ -277,15 +277,14 @@ For example this definition of Tempest object:
 will ensure that tempest will be executed with tempest.conf that looks like this:
 
 
-.. code-block:: toml
```
lpiwowar, line 280: I do not know why but the job started to fail. It was complaining about the toml formatting.

kopecmartin (approved):
lgtm

PR #30, opened 2024-01-24
Ensure controllers can access persistentvolumeclaims

lpiwowar:
This patch ensures that the controllers can access the persistent volume claims by updating the comments for kubebuilder.


kopecmartin (approved):

lpiwowar:
@kopecmartin, thanks!

PR #32, opened 2024-01-24
Change the interface for the tobiko CRD

lpiwowar:
With this change in the tobiko image we have to change the tobiko CRD. We will no longer use ENV variables to specify individual config options in tobiko.conf. Instead we are letting the user to specify the tobiko.conf directly in the Tobiko CRD [1].

This patch also introduces two new options (privateKey and publicKey) to comply with the tobiko tcib image [2].

[1] https://github.com/openstack-k8s-operators/tcib/pull/120
[2] https://github.com/openstack-k8s-operators/tcib/pull/120


kopecmartin (approved):
lgtm

PR #33, opened 2024-01-26
Allow the user to specify storageclass

lpiwowar:
This patch makes sure that the user can specify a storageclass that is used by the test-operator to generate pvc. Previously the test-operator could fail when there was no default storageclass.

### DIFF
--- a/config/crd/bases/test.openstack.org_tobikoes.yaml
+++ b/config/crd/bases/test.openstack.org_tobikoes.yaml
@@ -65,6 +65,10 @@ spec:
                 default: ""
                 description: Public Key
                 type: string
+              storageClass:
+                default: local-storage
+                description: StorageClass used to create PVCs that store the logs
+                type: string
               testenv:
                 default: py3
                 description: Test environment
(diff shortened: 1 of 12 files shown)


config/crd/bases/test.openstack.org_tobikoes.yaml
```diff
@@ -73,6 +77,8 @@ spec:
                 default: ""
                 description: Tobiko version
                 type: string
+            required:
```
afazekas, line 80: storage-class should not be required.

ATM the only requirement is being writable AFAIK.

controllers/common.go
```diff
@@ -61,6 +67,7 @@ func (r *Reconciler) EnsureLogsPVCExists(ctx context.Context, instance client.Ob
 					corev1.ResourceStorage: k8sresource.MustParse("1Gi"),
 				},
 			},
+			StorageClassName: &StorageClassName,
```
afazekas, line 70: Did you considered adding the vole claim parameters as struct to the parameters ?

afazekas (commented):
Some notes inside.

config/crd/bases/test.openstack.org_tobikoes.yaml
lpiwowar: StorageClass is required when there is no default storage class present in the environment. Or am I missing something?

PR #35, opened 2024-01-29
[docs] Update the "Getting Logs" section and Tobiko CRD

lpiwowar:
Based on the feedback from the users of the documentation we need to make the "Getting Logs" section more readable. Specifically, we want to:
- Remind the readers that by removing the CR you also remove the PV containing the logs.
- Be more explicit in the fact that there are two options for how to perform the second step of getting the logs.

This PR also updates the tobiko CR so that it contains the new values: privateKey and publicKey.


omersch381:
Looks good to me

kopecmartin (approved):
lgtm

PR #37, opened 2024-02-01
[docs] Update running tempest via pod

lpiwowar:
This patch updates the "Run tempest in a pod" section. This change makes sure that the tls certificate is properly mounted into the tempest pod.


lpiwowar:
Before merging this, let's ensure that this change works on every environment.

kopecmartin (approved):

PR #41, opened 2024-02-07
Add test-operator period job

lpiwowar:
Let's try to define daily periodic job that will execute tempest tests via the test-operator.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,0 +1,47 @@
+---
+- project:
+    name: openstack-k8s-operators/test-operator
+    periodic:
+      jobs:
+        - podified-multinode-edpm-deployment-crc:
+            vars:
+              test_fw: test_operator
+              cifmw_test_operator_concurrency: 4
+              cifmw_test_operator_timeout: 7200
+              cifmw_test_operator_tempest_include_list: |
+                tempest.
+              cifmw_test_operator_tempest_exclude_list: |
+                tempest.api.compute.admin.test_live_migration.LiveMigrationTest.test_live_migration_with_trunk
+                tempest.api.compute.admin.test_auto_allocate_network.AutoAllocateNetworkTest.test_server_multi_create_auto_allocate
+                tempest.api.compute.servers.test_server_rescue.ServerStableDeviceRescueTestIDE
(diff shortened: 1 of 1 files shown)


.zuul.yaml
```diff
@@ -0,0 +1,30 @@
+---
+job:
+  - test-operator-podified-multinode-edpm-deployment-crc:
```
queria, line 3: this job does not exist yet

i'm not 100% sure, but I believe You have to 'define' it here (so instead of this line) like:
```
- job:
    name: test-operator-podified-multinode-edpm-deployment-crc
    parent: podified-multinode-edpm-deployment-crc
    dependencies: ["openstack-k8s-operators-content-provider"]
    vars:
        ...
    required-projects: ...
```

.zuul.yaml
```diff
@@ -0,0 +1,27 @@
+---
+- project:
+    name: openstack-k8s-operators/test-operator
+    github-check:
+      jobs:
+        - podified-multinode-edpm-deployment-crc:
+            vars:
+              test_fw: test_operator
```
kopecmartin, line 8: let's set this var:
cifmw_test_operator_tempest_include_list
and run all tests? so that we see what fails

.zuul.yaml
```diff
@@ -0,0 +1,27 @@
+---
+- project:
+    name: openstack-k8s-operators/test-operator
+    github-check:
```
kopecmartin, line 4: periodic?

kopecmartin (commented):
works like a charm:
https://logserver.rdoproject.org/41/41/8b8fd46ceaf2662f217ae550f35631a5063fedcb/github-check/podified-multinode-edpm-deployment-crc/d4487da/controller/ci-framework-data/tests/test_operator/stestr_results.html

.zuul.yaml
lpiwowar: Yes, true. The github-check pipeline was used in order to be able to test the job and see whether zuul works correctly in the test-operator repo.

lpiwowar:
recheck

lpiwowar:
recheck

kopecmartin (approved):
nice work! lgtm

PR #42, opened 2024-02-13
Run tobiko pod as root with NET_ADMIN,NET_RAW capabilities

lpiwowar:
Tobiko requires to run as root in order to be able to run tcpdump command and needs NET_ADMIN and NET_RAW capabilities. Tcpdump command is currently being used by only one test in tobiko:

- tobiko/tests/scenario/neutron/test_qos.py::QoSNetworkTest::test_ping_dscp

### DIFF
--- a/pkg/tobiko/job.go
+++ b/pkg/tobiko/job.go
@@ -18,8 +18,8 @@ func Job(
 	envVars map[string]env.Setter,
 ) *batchv1.Job {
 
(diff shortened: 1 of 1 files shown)


kopecmartin (approved):
looks reasonable to me, i'll wait with merge for Edu's review

eduolivares (approved):

PR #43, opened 2024-02-15
Enable hostNetwork on test pods

eduolivares:
In order to avoid connectivity problems from the test-operator, this patch enables hostNetwork on both tempest and tobiko pods by default. Hence, the test pods will use the OCP worker (or CRC) network configuration directly.

### DIFF
--- a/pkg/tempest/job.go
+++ b/pkg/tempest/job.go
@@ -31,6 +31,7 @@ func Job(
 			BackoffLimit: instance.Spec.BackoffLimit,
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
+					HostNetwork:        true,
 					RestartPolicy:      corev1.RestartPolicyNever,
 					ServiceAccountName: instance.RbacResourceName(),
 					SecurityContext: &corev1.PodSecurityContext{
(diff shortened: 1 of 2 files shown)


lpiwowar (approved):
I've just tested the change locally and it does not seem to break anything that was already working. I think we should go for it. If it breaks something we can always revert.

lpiwowar:
I'm not going to merge it immedietally. I will give time @slawqo to do the review as well.

slawqo:
Why we need to have access to host networking? Wouldn't be e.g. access to the tenant or control plane network enough? You can connect POD to those kinds of networks as You want. See for the "networkAttachments" parameter in e.g. https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml#L225

lpiwowar:
@slawqo, @eduolivares what is the update on this PR, please? Can we close this or is the enabled hostNetwork still required for BGP?

eduolivares:
@lpiwowar @slawqo 
I am working now on the creation of the d/s BGP job:
https://issues.redhat.com/browse/OSPRH-2146

I think it makes sense to postpone the discussion about how test-operator will need to be configured to run tests on BGP setups until that task is completed.
So, I'm closing this PR and eventually will open it again or create a new one.

PR #45, opened 2024-02-16
[Tobiko] Add CAP_AUDIT_WRITE capability

lpiwowar:
The tobiko image uses sudo command that requires a CAP_AUDIT_WRITE capabilities in order to be able to write to kernel audit log. If the capability is missing we get the following error msg:

"sudo: unable to send audit message: Operation not permitted"

The message has no influence on successful execution of the tests but can be confusing for the users of the test-operator. Therefore we are enabling the capability by default.


eduolivares (approved):

kopecmartin (approved):
lgtm

PR #46, opened 2024-02-16
DNM testing srbac

kopecmartin:
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1159

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,7 +1,7 @@
 ---
 - project:
     name: openstack-k8s-operators/test-operator
-    periodic:
+    github-check:
       jobs:
         - podified-multinode-edpm-deployment-crc: &podified_multinode_edpm_deployment_crc
             vars:


kopecmartin:
the options were set in tempest.conf as expected - https://logserver.rdoproject.org/46/46/605e4aabbf174c084d65dbd8f24464167cd7df8a/github-check/podified-multinode-edpm-deployment-crc/82f7d45/controller/ci-framework-data/tests/test_operator/etc/tempest.conf .. now we need to verify if those are all the options we want to set by default

kopecmartin:
> the options were set in tempest.conf as expected - https://logserver.rdoproject.org/46/46/605e4aabbf174c084d65dbd8f24464167cd7df8a/github-check/podified-multinode-edpm-deployment-crc/82f7d45/controller/ci-framework-data/tests/test_operator/etc/tempest.conf .. now we need to verify if those are all the options we want to set by default

when I compare that with the results of the periodic run (without the srbac settings) the passing rate is the same .. that begs the question if we enabled really all srbac settings we want/require

kopecmartin:
recheck

kopecmartin:
recheck

PR #50, opened 2024-02-21
Add relevant tobiko env variables

eduolivares:
test-operator should pass the following env variables to the tobiko test pod:
- TOBIKO_TESTENV (was implemented before this patch)
- TOBIKO_VERSION (was implemented, but the env variable was not set)
- TOBIKO_PYTEST_ADDOPTS
- TOBIKO_PREVENT_CREATE
- TOX_NUM_PROCESSES

With these patch, the previous variables can optionally be defined in the tobiko CR.

OSPRH-5064


eduolivares:
Tested locally with successful results using the following values:
```
  preventCreate: true
  pytestAddopts: >-
   --skipregex='test_get_neutron_client_with_not_shared'
  version: osp-16.2

```

controllers/tobiko_controller.go
```diff
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+    "strconv"
```
lpiwowar, line 21: super small nit ... The tabs :D ... I understand.

lpiwowar (approved):
Looks good to me!

controllers/tobiko_controller.go
eduolivares: no, it's fine. I'm fixing it.
Thanks for the catch :)

lpiwowar:
@eduolivares thank you!

PR #52, opened 2024-02-26
Fix/docs

lpiwowar:
Fix deployerInput, profile, accounts sections and deployerInput format

This patch fixes:

- small inaccuracies in the TempestCR in the documentation.
- and file extension for the deployerInput file created by the test operator. The correct extension is .ini not .yaml


kopecmartin (approved):
lgtm

PR #54, opened 2024-03-01
[Tobiko] Workflows

lpiwowar:
This patch introduces workflows to the Tobiko CR. The workflows allow
to specify multiple executions of tobiko in a single CR. The steps
for a workflow are specified in a newly introduced Workflow section.
Values specified in the Workflow section have a higher priority and
overwrite configuration in the Spec section.

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/146


controllers/tobiko_controller.go
```diff
@@ -288,3 +275,67 @@ func (r *TobikoReconciler) EnsureTobikoCloudsYAML(ctx context.Context, instance
...
+//     that should be executed.
+func (r *TobikoReconciler) PrepareTobikoEnvVars(ctx context.Context, instance *testv1beta1.Tobiko, helper *helper.Helper, step int) map[string]env.Setter {
+	// Prepare env vars
+	envVars := make(map[string]env.Setter)
+	envVars["TOBIKO_TESTENV"] = env.SetValue(r.OverwriteStringValueWithWorkflow(instance.Spec.Testenv, instance.Spec.Workflow[step].Testenv, "py3"))
+	envVars["TOBIKO_VERSION"] = env.SetValue(r.OverwriteStringValueWithWorkflow(instance.Spec.Version, instance.Spec.Workflow[step].Version, ""))
+	envVars["TOBIKO_PYTEST_ADDOPTS"] = env.SetValue(r.OverwriteStringValueWithWorkflow(instance.Spec.PytestAddopts, instance.Spec.Workflow[step].PytestAddopts, ""))
+	envVars["USE_EXTERNAL_FILES"] = env.SetValue("True")
+
+	if instance.Spec.PreventCreate {
```
eduolivares, line 291: don't we need something like this here?
`r.OverwriteBoolValueWithWorkflow(instance.Spec.PreventCreate, instance.Spec.Workflow[step].PreventCreate, false)`

lpiwowar: True, thank you. I'm going to fix this.

controllers/common.go
```diff
@@ -22,13 +23,43 @@ import (
...
+		return ""
+	}
+}
+
+func (r *Reconciler) GetWorkflowConfigMapName(instance client.Object) string {
+	return instance.GetName() + workflowNameSuffix
+}
+
+func (r *Reconciler) GetLogDirName(frameworkName string, instance client.Object, workflowStepNum int) string {
+	return frameworkName + "-" + instance.GetName() + logDirNameInfix + strconv.Itoa(workflowStepNum)
```
eduolivares, line 60: IIUC, the log dir name could be something like this:
"<someprefix>-<workflowstepname>-workflow-step-<workflowStepNum>-workflow-step-<workflowStepNum>"

Aren't the last elements redundant?

eduolivares (approved):
Apart from my comment about log dir names, the PR looks good to me.
It would be good if @slawqo can take a look at it as well.

lpiwowar:
I'm going to merge this. The PR seems to be working. After merging this I'm going to immediatelly check whether the test-operator job got broken (if yes, I'm going to revert the change). I'm going to keep the discussion about the dir name open. This can always be fixed without any serious problems.

PR #60, opened 2024-03-12
[Tempest] Add correct labels

lpiwowar:
With this update [1] in the test_operator ci-framework role we are no longer waiting for a job using its name but labels named "workflowStep" and "instanceName". This patch adds these labels for tempest in order to prevent the job timeouts.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1235


kopecmartin (approved):
lgtm

PR #62, opened 2024-03-14
[Tempest] Introduce KubeconfigSecretName parameter

lpiwowar:
This patch introduces KubeconfigSecretName parameter that can be used to mount kubeconfig to the tobiko test pod. The parameter accepts a name of a secret that contains the content of kubeconfig under "config" key value. The kubeconfig is mounted to
/var/lib/tempest/.kube/config

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/155


lpiwowar:
It looks like this change is not needed after all. We can always reopen it. I'm closing it for now to cleanup the PRs.

PR #64, opened 2024-03-20
Skip identity tests

lpiwowar:
A couple of identity tests are failing due to know bug in keystone. Let's skip those for now as we want to see the periodic job failing only when there is something wrong with the test-operator.

Example of the failure here -> https://logserver.rdoproject.org/periodic/github.com/openstack-k8s-operators/test-operator/main/podified-multinode-edpm-deployment-crc/638fc04/controller/ci-framework-data/tests/test_operator/stestr_results.html


kopecmartin (approved):
lgtm

PR #68, opened 2024-03-27
Add variables for extra image creation

kstrenkova:
This patch adds variables that work with tcib on adding extra image for advanced test in Whitebox Neutron Tempest plugin. By setting the new variable WhiteboxNeutronExtraImage, tcib downloads specified image and sets the right confinguration options to work with image for advaned tests.

### DIFF
--- a/api/v1beta1/tempest_types.go
+++ b/api/v1beta1/tempest_types.go
@@ -101,6 +101,10 @@ type TempestRunSpec struct {
         // If this option is specified then only tests that are part of
         // the external plugin can be executed.
         ExternalPlugin []ExternalPluginType `json:"externalPlugin,omitempty"`
+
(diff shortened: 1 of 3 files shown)


api/v1beta1/tempest_types.go
```diff
@@ -101,6 +101,10 @@ type TempestRunSpec struct {
         // If this option is specified then only tests that are part of
         // the external plugin can be executed.
         ExternalPlugin []ExternalPluginType `json:"externalPlugin,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Add extra image for Whitebox Neutron Tempest plugin
```
lpiwowar, line 106: Just a small thing. Can you add this, please?
```
// +kubebuilder:default:=""
```

PR #69, opened 2024-04-02
[Tempest] Fix private key permissions

lpiwowar:
This patch changes the mount destination for the private key. The reason for this change is to make it possible to change the permissions for the private key to 600.


kopecmartin (approved):
feels like a hack but lgtm

lpiwowar:
I'm waiting for a tempest image that contains [this](https://github.com/openstack-k8s-operators/tcib/pull/159) change. If I did not wait I could break some jobs.

- https://quay.io/repository/podified-antelope-centos9/openstack-tempest?tab=tags&tag=latest

kopecmartin:
the depends on is merged

PR #74, opened 2024-04-08
Fix default value for storage class

lpiwowar:
The storageClass option is missing a default value. Thanks to this the pvc sometimes does not get created if it is running an an environment that does not have a default storage class set.


slawqo (approved):
thx

PR #75, opened 2024-04-09
Add variables for extra image creation

lpiwowar:
This patch adds variables that work with tcib on adding extra image for advanced test in Whitebox Neutron Tempest plugin. By setting the new variable WhiteboxNeutronExtraImage, tcib downloads specified image and sets the right confinguration options to work with image for advaned tests.


kopecmartin (approved):
lgtm

lpiwowar (commented):
LGTM, thank you @kstrenkova!

PR #76, opened 2024-04-10
Fix documentation deployment

jakubskunda:
In previous version of file documentation.yaml documentation page is deployed only when there was change in docs/ file. This patch will fix this problem.


kopecmartin (approved):
lgtm

PR #78, opened 2024-04-15
Add debug mode for test-operator

kstrenkova:
This patch introduces an option for test-operator that keeps the pod running even after a failure occured or after the file run_tempest.sh has reached its end (sleep infinity). It resolves the issue that oc rsh session is killed once the execution of the run_tempest.sh finishes.

### DIFF
--- a/api/v1beta1/tempest_types.go
+++ b/api/v1beta1/tempest_types.go
@@ -277,6 +277,11 @@ type TempestSpec struct {
 	Parallel bool `json:"parallel,omitempty"`
 
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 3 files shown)


lpiwowar:
Just a thought:). I'll need to test this manually but I was thinking that we might want to enable the debug mode only when the `workflow:` is not specified.

Otherwise it looks really good!:)

lpiwowar:
Once this change is ready for a review please rebase this PR on main so that the check job is triggered:).

lpiwowar:
@kstrenkova I assume this PR can be closed, right? :see_no_evil:

PR #80, opened 2024-04-19
Fix job being recreated

lpiwowar:
Let's change the created jobs by the test-operator to persistent ones. This prevents setting of TTL on jobs created by the test-operator and resolves the issue of jobs being recreated.


kopecmartin (approved):
lgtm, I like simple solutions

PR #85, opened 2024-04-23
Fix PVC creation

lpiwowar:
There was an issue of PVCs being stuck in Terminated stated when either Tempest CR or Tobiko CR was deleted and then quickly recreated. The issue is most likely caused by the fact the OSP needs some time to Terminate the PVC properly. The test-operator creates the PVC with a name based on the value stored in the name field in the metadata section. So if a CR is quickly reaplied it might lead to a situation when the test-operator is trying to create a PVC that at the same time is being terminated.

This patch adds a random suffix to the name of the PVC created by the test-operator that ensures that the test-operator creates a new PVC with each newly applied CR and leaves the old PVC to be terminated successfully.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1549


lpiwowar:
recheck

lpiwowar:
This is marked as a draft to protect this repo from accidentally merging this PR before the one in ci-framework is merged (https://github.com/openstack-k8s-operators/ci-framework/pull/1549).

lpiwowar:
recheck

kopecmartin (approved):
looks good, the check job passed

PR #86, opened 2024-04-25
Add correct index for the periodic job

lpiwowar:
Currently, the periodic job tries to build an test operator index that is connected to a specific PR. This does not work as there is no PR associated to a periodic job. This patch makes sure that the periodic job uses generic test-operator-index.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -9,6 +9,8 @@
         - podified-multinode-edpm-deployment-crc-test-operator:
             dependencies:
               - openstack-k8s-operators-content-provider
+            vars:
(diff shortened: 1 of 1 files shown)


kopecmartin (approved):

kopecmartin:
OMG! this super amazing github flow I love so much allowed me to merge this before the jobs were finished. Ah

lpiwowar:
Yes, I feel you :D. Zuul was down and because of that the job was aborted -> https://review.rdoproject.org/zuul/build/37f654cd82ba4adc8bcac3c5ee1283db

PR #92, opened 2024-05-10
Add debug mode and tempest webhook

kstrenkova:
This patch introduces an option for test-operator that keeps the pod running even after a failure occured or after the file run_tempest.sh has reached its end (sleep infinity). It resolves the issue that oc rsh session is killed once the execution of the run_tempest.sh finishes.

Additionally, it creates tempest webhook to check workflow option is not set when debug mode is set to true. It is to avoid unwanted behavior when in debug mode.

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/163


lpiwowar:
@kstrenkova, nice progress! We're going to utilize the webhook to check whether there are available PVs on the system later. This is going to be really helpful once completed. It looks like we are really close. We just need to resolve the TLS issue.

lpiwowar:
@kstrenkova you can add 
`Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/163`  into the commit message and description of this PR

.zuul.yaml
```diff
@@ -57,12 +57,6 @@
           identity.v3_endpoint_type public
           service_available.swift false
           service_available.cinder false
-      cifmw_test_operator_tempest_workflow:
```
lpiwowar, line 60: I know we removed it just for the testing purposes:). But this cannot be removed as part of this PR.

api/v1beta1/tempest_types.go
```diff
@@ -177,7 +177,7 @@ type TempestRunSpec struct {
 	// +kubebuilder:validation:Optional
 	// Extra images that should be downloaded inside the test pod and uploaded to
 	// openstack.
-	ExtraImages []ExtraImagesType `json:"extraImages"`
+	ExtraImages []ExtraImagesType `json:"extraImages,omitempty"`
```
lpiwowar, line 180: This is needed otherwise the values will be set to null on the side of the webhook. It's nice this is fixed!

api/v1beta1/tempest_types.go
```diff
@@ -351,6 +351,11 @@ type TempestSpec struct {
 	// spawned by the test operator.
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
+	// +kubebuilder:validation:Optional
+        // +kubebuilder:default=false
+        // Activate debug mode
```
lpiwowar, line 356: Can the explanation of this parameter be a little bit expanded? Maybe something like this (?):

```
Activate debug mode. When debug mode is activated any error encountered inside 
the test-pod causes that the pod will be kept alive indefinitely (stuck in "Running" phase) 
or until the corresponding Tempest CR is deleted This allows to debug any potential troubles
with `oc rsh`.
```

Maybe you'll find a better way how to describe this feature:).

lpiwowar: Also do not forget to run `make generate && make manifests` after you update the string here:)

lpiwowar:
@kstrenkova can you also please add the commit that updates the documentation please?

.zuul.yaml
kstrenkova: You are right, I have returned the test for this PR.

api/v1beta1/tempest_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	if len(r.Spec.Workflow) > 0 {
+		workflowActive = true
+	}
+
+        debugActive := r.Spec.Debug
+
+        if workflowActive && debugActive {
+            return nil, errors.New("Workflow variable must be empty to run debug mode!")
+        }
+	return nil, nil
```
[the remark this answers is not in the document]
kstrenkova: This code indeed looks nicer! I forgot to double-check code syntax after celebrating that it works. I have changed it accordingly.

api/v1beta1/tempest_types.go
kstrenkova: Very nicely written, I have changed the description and also ran the `make generate && make manifests` commands.

.zuul.yaml
lpiwowar: Thanks!:)

api/v1beta1/tempest_webhook.go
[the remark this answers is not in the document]
lpiwowar: I understand:). Thanks for the fix!:)

api/v1beta1/tempest_types.go
lpiwowar: Thank you!:)

lpiwowar:
recheck
The jobs got DEQUEUED.

docs/source/FAQ.rst
```diff
@@ -145,3 +145,31 @@ The issue can be fixed in two steps:
...
+    $ make docker-build IMG=<registry>/<user>/<operator_image_name>:<tag>
+    $ make docker-push IMG=<registry>/<user>/<operator_image_name>:<tag>
+
+2. Create test-operator-bundle image
+
+.. code-block:: bash
+
+    $ make bundle IMG=<registry>/<user>/<operator_image_name>:<tag>
+    $ make bundle-build BUNDLE_IMG=<registry>/<user>/<bundle_image_name>:<tag>
+    $ make docker-push IMG=<registry>/<user>/<bundle_image_name>:<tag>
```
lpiwowar, line 167: This is a small thing. Can you please double check whether the `make bundle-push` really doesn't work for you? I've tested the steps locally and this worked for me:

```
make bundle-push BUNDLE_IMG=<registry>/<user>/<bundle_image_name>:<tag>
```

It would be nice if we were consistent with the openshift documentation. You were right that docker-push here works as well:) :+1:

lpiwowar (changes requested):
Overall looks good to me! :+1: 

I've just added two small requests for the documentation.

docs/source/FAQ.rst
```diff
@@ -145,3 +145,31 @@ The issue can be fixed in two steps:
    free the PVs and you can continue with the testing but please make sure that
    you can afford to lose the data stored on the PVs you are about the free.
 
+
+**7. How can I build test-operator-index locally?**
```
lpiwowar, line 149: One more thing please :see_no_evil: :smile: . Can you modify this [1] line in the documentation:
```
ENABLE_WEBHOOKS=false make install run
```

It'll make the installation easier for lot of people (most of the people won't need to modify the webhooks during development). We can extend the documentation later.

[1] https://github.com/openstack-k8s-operators/test-operator/blob/main/docs/source/guide.rst#running-test-operator-locally-outside-the-cluster

lpiwowar (changes requested):

docs/source/FAQ.rst
kstrenkova: I have added the changes, please check if they are as you requested :smile:

lpiwowar: The changes look good:) ... Thank you!

lpiwowar:
Let's wait for the TCIB change to be merged first:). Also we should wait for the latest build of the openstack-tempest[-all] image that contains the @kstrenkova's change [1].

[1] https://quay.io/repository/podified-antelope-centos9/openstack-tempest-all?tab=tags

lpiwowar (approved):

lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar (approved):
LGTM! :tada:

PR #95, opened 2024-05-15
[Tempest][Tobiko] Add nodeSelector and Tolerations 

lpiwowar:
This patch adds support for specifying tolerations and nodeSelector
in both the Tempest and the Tobiko CR.


config/crd/bases/test.openstack.org_tobikoes.yaml
```diff
@@ -93,6 +99,47 @@ spec:
                 default: py3
                 description: Test environment
                 type: string
+              toleration:
```
eduolivares, line 102: should it be `toleration` or `tolerations`?
Is it important?

lpiwowar: Yes, I think it is important. Let me fix it:). The change would probably work but the name of the parameter would be inconsistent with the tempest one.

eduolivares (approved):

PR #96, opened 2024-05-16
Use RWX PVs only for parallel execution

lpiwowar:
This change is needed as we are going to switch from local storage to LVMS based storage that does not support RWX. Because of that the test-operator will now be by default requesting RWO PVs only. RWX will be requested when parallel is set to true in the test-operator related CR.


kopecmartin (approved):
seems good

PR #102, opened 2024-06-06
[Tempest] Install extra RPMs

lpiwowar:
This patch introduces tempestRun.extraRPMs parameter in the Tempest CR that allows to specifying a list of extra RPMs that should be installed before the execution of Tempest.

This parameter accepts a list of URLs that point to the RPMs that should be installed. The RPMs are installed in the order in which they are listed via the parameter.

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/175

### DIFF
--- a/api/bases/test.openstack.org_tempests.yaml
+++ b/api/bases/test.openstack.org_tempests.yaml
@@ -211,6 +211,12 @@ spec:
                       - name
                       type: object
                     type: array
+                  extraRPMs:
+                    description: A list URLs that point to RPMs that should be downloaded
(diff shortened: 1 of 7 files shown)


lpiwowar:
This can be merged once this PR is merged -> https://github.com/openstack-k8s-operators/tcib/pull/175

kopecmartin:
looks awesome! add please also some doc and a warning it can't be used together with the extra plugins opt

kopecmartin (approved):
lgtm

PR #103, opened 2024-06-10
update csv for recent changes

yazug:
recent pull request injected this discrepancy

this patch was generated with make bundle

### DIFF
--- a/config/manifests/bases/test-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/test-operator.clusterserviceversion.yaml
@@ -15,6 +15,59 @@ spec:
       displayName: Tempest
       kind: Tempest
       name: tempests.test.openstack.org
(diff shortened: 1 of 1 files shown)


lpiwowar (commented):
Thank you @yazug! We've just merged change that updates the CSV as well here [1]. 

I think we can close this PR but what we should do is to extend the CSV so that all the parameters are properly described (e.g. [2], [3]).

[1] https://github.com/openstack-k8s-operators/test-operator/pull/92
[2] https://github.com/openstack-k8s-operators/test-operator/blob/de723c7009e9d7406fd46cdf788dcb11c9251353/api/v1beta1/tempest_types.go#L379
[3] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/manifests/bases/openstack-operator.clusterserviceversion.yaml

PR #105, opened 2024-06-14
Change name of image env variables

lpiwowar:
The name of the env variables specifies name for variables in the openstackcontrolplane CR. It is more accurate to prefix these vars with Test to clearly indicate that these variables contain URLs that point to images that should be used for testing only.

### DIFF
--- a/config/default/manager_default_images.yaml
+++ b/config/default/manager_default_images.yaml
@@ -11,7 +11,7 @@ spec:
       containers:
       - name: manager
         env:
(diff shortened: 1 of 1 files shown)


lpiwowar:
recheck
It looks like there's something wrong with tobiko.

kopecmartin (approved):
lgtm

lpiwowar:
recheck
The tobiko job was fixed with this PR -> https://review.opendev.org/c/x/tobiko/+/922184

PR #107, opened 2024-06-18
Add --create when tempestConf section is empty

kstrenkova:
When tempestConf section is empty in Tempest CR the default value for create is set to false, which leads to discover-tempest-config failure, even though the CR is valid. This patch sets create to true if the tempestConf section is empty to prevent the failure.

Additionally, this patch corrects local configuration of webhooks as this patch works with mutating webhook and needs these changes to work properly.

### DIFF
--- a/api/v1beta1/tempest_webhook.go
+++ b/api/v1beta1/tempest_webhook.go
@@ -63,6 +63,10 @@ func (spec *TempestSpec) Default() {
         if spec.ContainerImage == "" {
 		spec.ContainerImage = tempestDefaults.ContainerImageURL
 	}
+
+	if spec.TempestconfRun == (TempestconfRunSpec{}) {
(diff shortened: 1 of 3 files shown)


lpiwowar (commented):
LGTM! :+1: 
I've tested the change and it works as expected! Nice work :tada: 

We just have to deal now with the failing check job. Can you please rebase this patch on top of main. There was a recent change in `.zuul.yaml` that should be fixing the issue [1].

[1] https://github.com/openstack-k8s-operators/test-operator/pull/110

lpiwowar:
recheck

I want to see whether the job will pass when the PR is proposed from a fork.

lpiwowar:
@kopecmartin if you agree we can merge this as well.

lpiwowar (approved):
LGTM! :+1:

PR #108, opened 2024-06-18
Add correct labels

lpiwowar:
This patch ensures that resources related to the test-operator are created with the `openstack.org/operator-name: test-operator` label. This fixes the webhook issue as the service created for the webhook was matching all the operators with
the `control-plane: controller-manager` label.


lpiwowar:
@kopecmartin If you agree I would merge this.

kopecmartin (approved):

PR #110, opened 2024-06-18
[zuul] Disable tobiko testing

lpiwowar:
Let's disable tobiko testing in the check job for now as there is a numpy
related bug in tobiko that should be fixed with this PR [1]. We can
enable tobiko testing again once the change fully propagates into the
tobiko image.

[1] https://review.opendev.org/c/x/tobiko/+/922184


kopecmartin (approved):
lgtm, we're blocked by the tobiko issue

PR #113, opened 2024-06-19
Print warning message for update of existing CR

kstrenkova:
As of now, the Tempest CR does not support direct updates of already existing CR instance. This patch adds warning message when user tries to update an already existing instance of a CR. It helps to prevent using unsupported update of CR and problems tied with this misuse.

### DIFF
--- a/api/v1beta1/tempest_webhook.go
+++ b/api/v1beta1/tempest_webhook.go
@@ -24,6 +24,9 @@ package v1beta1
 
 import (
         "errors"
+
+	"github.com/google/go-cmp/cmp"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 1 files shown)


api/v1beta1/tempest_webhook.go
```diff
@@ -84,8 +84,15 @@ func (r *Tempest) ValidateCreate() (admission.Warnings, error) {
...
 
-	// TODO(user): fill in your validation logic upon object update.
-	return nil, nil
+	oldTempest, ok := old.(*Tempest)
+	if !ok || oldTempest == nil {
+		return nil, errors.New("Unable to convert existing object.")
+	}
+
+	warnings := admission.Warnings{}
+	warnings = append(warnings, "You are updating an already existing instance of a Tempets CR! Be aware that changes won't be applied.")
```
lpiwowar, line 93: I'm sorry @kstrenkova for being inaccurate in the ticket description. We need to make this an error as we do not want to pass the update of the Tempest CR to the test-operator-controller.

kstrenkova:
@lpiwowar No worries, I tried to combine warning and error. Maybe you would prefer only error, please let me know. 

The new version of this patch uses library cmp to compare old and new Spec values, which seems to be working locally like it's supposed to. Although, I am not sure if this is a safe way of doing things as the documentation for cmp says "It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. ". But nova operator uses this library and I have also not found a better alternative yet. Let me know what you think :D

lpiwowar:
@kstrenkova I tested your PR and it works as expected! It seems that using `cmp` is the best option here. In nova-operator they're using it as well [1]. With that being I think you raised valid questions here. Thank you!

@kopecmartin can you please merge this? :pray:  The job is still failing because of the tobiko bug.

[1] https://github.com/openstack-k8s-operators/nova-operator/blob/ed8c931f68318a3142cc6d638abddeba7381cce1/api/v1beta1/novaapi_webhook.go#L119

lpiwowar (approved):
LGTM! :+1:

kopecmartin (approved):
thanks Katka!

PR #118, opened 2024-07-01
Request separate PVCs when parallel is set to true

kstrenkova:
This patch changes the way PVCs are requested when the parallel option is set to true. Instead of using one PV for all the workflow steps, it is divided into multiple PVs. This change is needed, because the LVM backed storage does not support the RWX, which causes issues in some cases.

### DIFF
--- a/controllers/tempest_controller.go
+++ b/controllers/tempest_controller.go
@@ -201,6 +201,13 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		"operator":         "test-operator",
 	}
 
+	workflowStepNum := 0
+
+	// Create multiple PVCs for parallel execution
+	if instance.Spec.Parallel && externalWorkflowCounter < len(instance.Spec.Workflow) {
+		workflowStepNum = externalWorkflowCounter
+	}
+
 	// Create PersistentVolumeClaim
 	ctrlResult, err := r.EnsureLogsPVCExists(
 		ctx,
@@ -208,7 +215,7 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		helper,
 		serviceLabels,
 		instance.Spec.StorageClass,
-		instance.Spec.Parallel,
+		workflowStepNum,
 	)
 
 	if err != nil {
@@ -268,7 +275,7 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	customDataConfigMapName := GetCustomDataConfigMapName(instance, externalWorkflowCounter)
 	EnvVarsConfigMapName := GetEnvVarsConfigMapName(instance, externalWorkflowCounter)
 	jobName := r.GetJobName(instance, externalWorkflowCounter)
-	logsPVCName := r.GetPVCLogsName(instance)
+	logsPVCName := r.GetPVCLogsName(instance, workflowStepNum)
 	containerImage := r.GetContainerImage(ctx, helper, instance.Spec.ContainerImage, instance)
 
 	// Note(lpiwowar): Remove all the workflow merge code to webhook once it is done.
--- a/controllers/ansibletest_controller.go
+++ b/controllers/ansibletest_controller.go
@@ -166,7 +166,7 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		helper,
 		serviceLabels,
 		instance.Spec.StorageClass,
-		false,
+		0,
 	)
 	if err != nil {
 		return ctrlResult, err
@@ -202,7 +202,7 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	mountCerts := r.CheckSecretExists(ctx, instance, "combined-ca-bundle")
 	jobName := r.GetJobName(instance, externalWorkflowCounter)
 	envVars, workflowOverrideParams := r.PrepareAnsibleEnv(ctx, serviceLabels, instance, helper, externalWorkflowCounter)
-	logsPVCName := r.GetPVCLogsName(instance)
+	logsPVCName := r.GetPVCLogsName(instance, 0)
 	containerImage := r.GetContainerImage(ctx, helper, workflowOverrideParams["ContainerImage"], instance)
 	jobDef := ansibletest.Job(
 		instance,
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -176,12 +176,13 @@ func (r *Reconciler) GetWorkflowConfigMapName(instance client.Object) string {
 	return instance.GetName() + workflowNameSuffix
 }
 
-func (r *Reconciler) GetPVCLogsName(instance client.Object) string {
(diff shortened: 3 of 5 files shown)


lpiwowar:
The job is failing. But this is expected as we need to update the other controllers as well (tobiko + horizontest).
```
  # github.com/openstack-k8s-operators/test-operator/controllers
    controllers/horizontest_controller.go:128:3: cannot use instance.Spec.Parallel (variable of type bool) as int value in argument to r.EnsureLogsPVCExists
    controllers/horizontest_controller.go:163:34: not enough arguments in call to r.GetPVCLogsName
            have (*"github.com/openstack-k8s-operators/test-operator/api/v1beta1".HorizonTest)
            want (client.Object, int)
    Error: building at STEP "RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; CGO_ENABLED=0  GO111MODULE=on go build ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go": while running runtime: exit status 1
```

controllers/tempest_controller.go
```diff
@@ -208,7 +208,7 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		helper,
 		serviceLabels,
 		instance.Spec.StorageClass,
-		instance.Spec.Parallel,
+		0, // default naming
```
lpiwowar, line 211: Can you please simply the code so that there are no two calls of the EnsureLogsPVCExists() function?

lpiwowar (commented):
Overall a like the rough logic of the patch. There are some things that need to be fixed but I know @kstrenkova is aware. Good job I think really good start when you take into consideration that you started to work on this PR on your own!:) 

Also, when you'll be working on this @kstrenkova do not be surprised that some changes might be needed for the horizontest CR as well (newly introduced couple of days back) but I'm pretty sure you'll be able to figure out what to do!:) This issue might occur if it turns out that a rebase is needed.

:warning: We can not merge this before we update the test-operator ci-framework role as well. I can do that once I'll be back.

kstrenkova:
I have addressed every issue @lpiwowar has pointed out. Now the code looks much cleaner, thank you for the review! 

The first patch contained just change of logic for Tempest CR, now I have added changes to other CRs as well. One extra change is adding a placeholder for Ansible CR, because I want to apply the same logic to all CRs and it will be easier to just code the logic behind the placeholder later.

lpiwowar:
recheck
Failure in the previous job run was not caused by the patch.

kopecmartin:
recheck

lpiwowar:
recheck

api/bases/test.openstack.org_ansibletests.yaml
```diff
@@ -115,6 +115,13 @@ spec:
                 description: OpenStackConfigSecret is the name of the Secret containing
                   the secure.yaml
                 type: string
+              parallel:
+                default: false
+                description: By default test-operator executes the test-pods sequentially
+                  if multiple instances of test-operator related CRs exist. If you
+                  want to turn off this behaviour then set this option to true. TODO(kstrenko)
+                  This is a placeholder, logic is not implemented
```
lpiwowar, line 123: I'm just checking. Is this still true?

api/v1beta1/ansibletest_types.go
```diff
@@ -55,6 +55,15 @@ type AnsibleTestSpec struct {
 	// StorageClass used to create PVCs that store the logs
 	StorageClass string `json:"storageClass"`
 
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// By default test-operator executes the test-pods sequentially if multiple
+	// instances of test-operator related CRs exist. If you want to turn off this
+	// behaviour then set this option to true.
+	// TODO(kstrenko) This is a placeholder, logic is not implemented
```
lpiwowar, line 64: ditto

config/crd/bases/test.openstack.org_ansibletests.yaml
```diff
@@ -115,6 +115,13 @@ spec:
                 description: OpenStackConfigSecret is the name of the Secret containing
                   the secure.yaml
                 type: string
+              parallel:
+                default: false
+                description: By default test-operator executes the test-pods sequentially
+                  if multiple instances of test-operator related CRs exist. If you
+                  want to turn off this behaviour then set this option to true. TODO(kstrenko)
+                  This is a placeholder, logic is not implemented
```
lpiwowar, line 123: ditto

api/bases/test.openstack.org_ansibletests.yaml
kstrenkova: This is added for efficiency -> as far as I know AnsibleTest CR does not have parallel option yet, but will have in future. Thanks to adding this placeholder we can use the same logic (for creating multiple PVCs) as in Tempest and Tobiko CR and not open a ticket on adding this later.

If we do not want to add placeholder I can edit the change as it is in Horizon CR, but when the parallel option will be added later, we need to change the logic for creating multiple PVCs separately.

Maybe I overexplained it, but let me know what you think:)

lpiwowar: > If we do not want to add placeholder I can edit the change as it is in Horizon CR, but when the parallel option will be added later, we need to change the logic for creating multiple PVCs separately.

As discussed, let's use the same approach as with Horizon CR. The danger of this current approach is that we are exposing a new parameter to the user that in the end does nothing. Thank you for the explanation @kstrenkova!

controllers/tempest_controller.go
```diff
@@ -268,7 +275,7 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	customDataConfigMapName := GetCustomDataConfigMapName(instance, externalWorkflowCounter)
 	EnvVarsConfigMapName := GetEnvVarsConfigMapName(instance, externalWorkflowCounter)
 	jobName := r.GetJobName(instance, externalWorkflowCounter)
-	logsPVCName := r.GetPVCLogsName(instance)
+	logsPVCName := r.GetPVCLogsName(instance, externalWorkflowCounter)
```
lpiwowar, line 278: I did not notice it previously but this logic needs to be probably slightly updated. It will not work in case there is `instance.Spec.Parallel == false` and `len(workflow) > 0`. What would happen if we replaced `externalWorkflowCounter` with `workflowStepNum`?

lpiwowar (changes requested):
There is minor update required:). We can talk about this next week.

controllers/tempest_controller.go
kstrenkova: Good question. I have tried both cases just to be sure and you are right, the `instance.Spec.Parallel == false` and `len(workflow) > 0` does not work. When I put `workflowStepNum` there instead, it works as expected, so I will do these changes with my next push.

lpiwowar:
recheck

lpiwowar (approved):
LGTM! :+1: Thank you!

PR #123, opened 2024-07-04
Remove explicit reference to upstream image

lpiwowar:
We should not be hardcoding references to upstream images in the CRs. Images that should be used by the operator should be listed in the manager_default_images.yaml.

### DIFF
--- a/api/bases/test.openstack.org_ansibletests.yaml
+++ b/api/bases/test.openstack.org_ansibletests.yaml
@@ -76,7 +76,7 @@ spec:
                   in the ansible pod
                 type: string
               containerImage:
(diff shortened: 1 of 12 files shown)


arxcruz:
Are other operators doing the same? I don't see a problem since we are overwriting it

lpiwowar:
@arxcruz I'm not sure whether the other operators have default values set directly in the CR definition. I believe the only values that are overriden are here -> https://github.com/openstack-k8s-operators/test-operator/blob/5656a7097845b6a1ac492b657d5b39f0c9a33817/config/default/manager_default_images.yaml#L14

kopecmartin (approved):

PR #127, opened 2024-07-08
Fix panic:interface conversion

kopecmartin:
This fixes the following error:
panic: interface conversion: interface {} is map[string]interface {}, not map[interface {}]interface {} [recovered]


kopecmartin:
tobiko ran successfully, the issue is fixed

PR #128, opened 2024-07-08
Empty commit to make place for the v0.1.0 tag

gibizer:
To be able to distinguish between pseudoversions from main and from
18.0.0-proposed we need to start tagging main with v0.1.0 while keeping
18.0.0-proposed using the v0.0.x tags. Naturally we would only need to
start tagging after the first commit arrives to main. But to avoid
forgetting the tagging at that point we add an empty commit and tag it
instead.


stuggi (approved):
/lgtm

kopecmartin:
lgtm, let me merge this later today, there is one or 2 commits we need to land in master and rebase the 18.0.0-proposed branch

PR #130, opened 2024-07-10
Update lib-common/modules/common

stuggi:
Related: [OSPRH-8379](https://issues.redhat.com//browse/OSPRH-8379)

PR #131, opened 2024-07-10
[18.0.0-proposed] Update lib-common/modules/common

stuggi:
Related: [OSPRH-8379](https://issues.redhat.com//browse/OSPRH-8379)


kopecmartin (commented):
shouldn't be these changes the same as in https://github.com/openstack-k8s-operators/test-operator/pull/130 ? or do we want different hashes for 18.0.0 proposed?

stuggi:
> shouldn't be these changes the same as in #130 ? or do we want different hashes for 18.0.0 proposed?

main branch maps to lib-common main and 18.0.0-proposed to the lib-common 18.0.0-proposed branch

PR #132, opened 2024-07-10
Add workflow section to documentation

kstrenkova:
This patch adds a description of the workflow section to the test-operator documentation. As workflow is a new feature that can be difficult to use for users, there is a need to describe it in a greater detail.


lpiwowar (approved):
LGTM! :+1:

PR #133, opened 2024-07-10
Fix wrong descriptions for Tobiko CR options

kstrenkova:
Some Tobiko CR options have wrongly copy-pasted descriptions. This patch adds correct descriptions to fix this issue.

### DIFF
--- a/api/bases/test.openstack.org_tobikoes.yaml
+++ b/api/bases/test.openstack.org_tobikoes.yaml
@@ -49,10 +49,6 @@ spec:
                 default: ""
                 description: Container image for tobiko
                 type: string
-              debug:
-                default: true
-                description: Run tests in parallel
-                type: boolean
               kubeconfigSecretName:
                 default: ""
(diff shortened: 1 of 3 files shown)


api/bases/test.openstack.org_tobikoes.yaml
```diff
@@ -51,7 +51,7 @@ spec:
                 type: string
               debug:
                 default: true
-                description: Run tests in parallel
+                description: Debug option
```
kopecmartin, line 54: maybe you could elaborate a bit more, in tempest's api definition we have much more detailed description:
https://github.com/openstack-k8s-operators/test-operator/blob/main/api/bases/test.openstack.org_tempests.yaml#L70

kstrenkova: I have removed the debug options as they are not used / not doing anything. I think adding them was most likely copy-paste error.

kopecmartin (approved):
lgtm

PR #135, opened 2024-07-11
Add useful set of printcolumns for CRs

kstrenkova:
To give user more detailed information about the custom resource, additional printcolumns are needed. This patch adds the 'status' and 'message' printcolumns, which are frequently used in other operators.

### DIFF
--- a/api/v1beta1/horizontest_types.go
+++ b/api/v1beta1/horizontest_types.go
@@ -150,6 +150,8 @@ type HorizonTestStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[0].status",description="Status"
+//+kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.conditions[0].message",description="Message"
 
 // HorizonTest is the Schema for the horizontests API
 type HorizonTest struct {
--- a/api/v1beta1/tempest_types.go
+++ b/api/v1beta1/tempest_types.go
@@ -495,6 +495,8 @@ type TempestStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[0].status",description="Status"
+//+kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.conditions[0].message",description="Message"
 
 // Tempest is the Schema for the tempests API
 type Tempest struct {
--- a/api/v1beta1/tobiko_types.go
+++ b/api/v1beta1/tobiko_types.go
@@ -254,6 +254,8 @@ type TobikoStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[0].status",description="Status"
+//+kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.conditions[0].message",description="Message"
 
 // Tobiko is the Schema for the tobikoes API
 type Tobiko struct {
--- a/config/samples/test_v1beta1_ansibletests.yaml
+++ b/config/samples/test_v1beta1_ansibletests.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: openstack
 spec:
   extraMounts:
-    - Name: some-configmap
+    - name: some-configmap
       subPath: this.conf
       mountPath: /var/conf
   debug: true
--- a/controllers/horizontest_controller.go
+++ b/controllers/horizontest_controller.go
@@ -66,7 +66,7 @@ func (r *HorizonTestReconciler) GetLogger(ctx context.Context) logr.Logger {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
-func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
 	Log := r.GetLogger(ctx)
 
 	// How much time should we wait before calling Reconcile loop when there is a failure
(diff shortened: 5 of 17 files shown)


lpiwowar:
recheck

kstrenkova:
For now I have made small changes like adding setting DeploymentReadyCondition to true, when there is no error and changing IsReady function to have only relevant conditions.

With @lpiwowar, we have talked also about adding some MarkTrue functions before certain returns, but I think that can be replaced by not updating conditions to status Unknown every reconcile loop. So that is what I did in this new patch. Now the ReadyCondition is no longer stuck in Unknown state. BUT I have encountered a problem when ReadyCondition is stuck in a False state because it's waiting for DeploymentCondition to be true. Don't know what is the cause, maybe we can look at it together.

config/samples/test_v1beta1_ansibletests.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: openstack
 spec:
   extraMounts:
-    - Name: some-configmap
+    - name: some-configmap
```
lpiwowar, line 9: Nice catch!:)

lpiwowar (commented):
I will finish the review tomorrow. Just couple of questions for now.

controllers/tobiko_controller.go
```diff
@@ -144,10 +194,10 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		"operator":         "test-operator",
 	}
 
-	result, err := r.EnsureTobikoCloudsYAML(ctx, instance, helper, serviceLabels)
+	yamlResult, err := r.EnsureTobikoCloudsYAML(ctx, instance, helper, serviceLabels)
```
lpiwowar, line 197: Usually you want to modify only the code that is relevant to PR or commit. Ideally this should have been a separate commit. But you can keep it here. It's nice fix. I'm just mentioning as something to keep in mind.

controllers/horizontest_controller.go
```diff
@@ -86,6 +86,55 @@ func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	defer func() {
+		condition.RestoreLastTransitionTimes(&instance.Status.Conditions, savedConditions)
+		if instance.Status.Conditions.IsUnknown(condition.ReadyCondition) {
+			instance.Status.Conditions.Set(
+				instance.Status.Conditions.Mirror(condition.ReadyCondition))
+		}
+		// update the overall status condition if service is ready
+		if instance.IsReady() {
+			instance.Status.Conditions.MarkTrue(condition.ReadyCondition, condition.ReadyMessage)
+		}
```
lpiwowar, line 110: This should ideally be moved on line 102. Before we check that the ReadyCondition is Unknown:
```
...
if instance.Status.Conditions.IsUnknown(condition.ReadyCondition) {
...
```
This probably applies to the other reconcile loops as well (horizon, ansible, ...).

controllers/horizontest_controller.go
```diff
@@ -60,7 +60,7 @@ type HorizonTestReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
-func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
```
lpiwowar, line 63: Was this change needed because of `golangci|operator-lint`? I'm just curious:). 

This could have been a separate commit together with the renaming of the variable `result` -> `yamlResult`. It's ok to keep it as is right now. Just something to keep in mind.

api/v1beta1/horizontest_types.go
```diff
@@ -173,6 +175,10 @@ func init() {
 	SchemeBuilder.Register(&HorizonTest{}, &HorizonTestList{})
 }
 
+func (instance HorizonTest) IsReady() bool {
+       return instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition)
```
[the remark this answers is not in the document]
kstrenkova: When we were discussing this task in the office I remember you saying InputReadyCondition is not relevant to us. But I could be wrong and in that case we can use the AllsubConditionsTrue function, because I was just leaving out the InputReadyCondition and putting other conditions into IsReady function.

lpiwowar (commented):
Another set of comments:) (only small things). I'm going to return to this in a moment. The last thing I want to do is to test this manually.

api/v1beta1/horizontest_types.go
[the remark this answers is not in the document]
lpiwowar: Yes, true. What we want to do is to remove the `InputReadyCondition` from the conditionList as well. And from any code that references it [1].

[1] https://github.com/openstack-k8s-operators/test-operator/pull/135/files#diff-763e3c6d751fb09d0fb537d31b313f6bb32dea43ee46573de067f97e09a477e2R122

controllers/horizontest_controller.go
kstrenkova: To keep code across all operators consistent I added the _err error, because it was needed in the PatchInstance function. And then I needed to add result as well, because named and unnamed type can't be combined. And that resulted in changing result to yamlResult :sweat_smile: But you are right, it would have been better as it's own commit. I can try to make it a different commit by playing with git.

api/v1beta1/horizontest_types.go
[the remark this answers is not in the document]
kstrenkova: Alright, I have deleted all of InputReadyCondition mentions.

api/v1beta1/tobiko_types.go
```diff
@@ -277,6 +279,11 @@ func init() {
 	SchemeBuilder.Register(&Tobiko{}, &TobikoList{})
 }
 
+func (instance Tobiko) IsReady() bool {
+       return instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) &&
```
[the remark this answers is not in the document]
kstrenkova: I have changed this for checking all subconditions. I have tested this on tempest CR and it did work, but maybe a double-check wouldn't hurt. I will get to it tomorrow.

controllers/horizontest_controller.go
kstrenkova: Ack, changed. :smile:

api/v1beta1/tempest_types.go
```diff
@@ -518,6 +520,10 @@ func init() {
 	SchemeBuilder.Register(&Tempest{}, &TempestList{})
 }
 
+func (instance Tempest) IsReady() bool {
+	return instance.Status.Conditions.AllSubConditionIsTrue()
+}
+
```
lpiwowar, line 526: I do not know whether this is just a WIP change but I suggest we replace every usage of `IsReady()` with `instance.Status.Conditions.AllSubConditionIsTrue()`

lpiwowar:
Thank you @kstrenkova

PR #146, opened 2024-07-26
Fix make [golangci|operator-lint]

lpiwowar:


controllers/ansibletest_controller.go
```diff
@@ -110,7 +110,7 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	runningJobName := r.GetJobName(instance, externalWorkflowCounter-1)
 	err = r.Client.Get(ctx, client.ObjectKey{Namespace: instance.GetNamespace(), Name: runningJobName}, runningAnsibleJob)
```
stuggi, line 111: is it ok to proceed if `err !=nil` ? should there be a check for if err != nil and maybe special handling for IsNotFound if it is ok to proceed?

controllers/common.go
```diff
@@ -240,7 +241,11 @@ func (r *Reconciler) ReleaseLock(ctx context.Context, instance client.Object) bo
 		},
 	}
 
-	r.Client.Delete(ctx, &cm)
+	err := r.Client.Delete(ctx, &cm)
+	if err != nil && k8s_errors.IsNotFound(err) {
+		return false
+	}
```
stuggi, line 247: if there is an error during delete, which is not IsNotFound, is it correct that we return true?

controllers/common.go
```diff
@@ -221,14 +221,15 @@ func (r *Reconciler) AcquireLock(ctx context.Context, instance client.Object, h
 						Namespace: instance.GetNamespace(),
 					},
 				}
-				configmap.EnsureConfigMaps(ctx, h, instance, cms, nil)
-				return true
+
+				err = configmap.EnsureConfigMaps(ctx, h, instance, cms, nil)
+				return err == nil
 			} else {
```
stuggi, line 227: could remove the else since the prior if has a return

controllers/common.go
lpiwowar: This should be fixed, thank you!:)

controllers/common.go
lpiwowar: True, thank you!:)

controllers/ansibletest_controller.go
lpiwowar: Yes, better processing of the err would be better here. Thank you! I've added an if-statement that deals with the error. `IsNotFound` error is acceptable in this stage of the control loop as the job gets created in a later stage of the control loop.

lpiwowar (commented):
Thank you for the review!

stuggi:
lgtm from the perspective on not knowing the internals of the operator.

PR #153, opened 2024-08-01
Add debug mode for Tobiko

kstrenkova:
Similar to debug mode for Tempest, this patch introduces new option for test-operator that keeps the pod running after a successful run or even after a failure has occured. It resolves the issue that oc rsh session is killed once the execution of the run_tobiko.sh file finishes.


lpiwowar:
We can merge this once the tcib change is merged as well -> [here|https://github.com/openstack-k8s-operators/tcib/pull/193]

lpiwowar (approved):
LGTM! Thank you @kstrenkova

PR #154, opened 2024-08-01
Protect lock from deletion by different reconcile loop

lpiwowar:
Previously, the logic for releasing test operator lock was written in such a way that it allowed Tempest reconcile loop to delete lock acquired by Tobiko reconcile loop.

This patch fixes this issue by introducing owner field for the test operator lock. The releaceLock function now checks whether the reconcile loop that is trying to delete the lock owns it before proceeding with deletion.


eduolivares (approved):
verified

eduolivares (approved):

PR #155, opened 2024-08-02
Fix make [golangci|operator-lint]

lpiwowar:


lpiwowar:
The change should be safe to merge as the check job passed.

PR #168, opened 2024-08-12
Refactor merging of workflow section

kstrenkova:
Currently, the merging of workflow section has a lot of duplicate code. We determine which section is used checking each parameter's value in the workflow. In the newly proposed code all resource types use the same way to overwrite main section with workflow section, which makes the code clearer and easier to add new parameters.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2906


controllers/tempest_controller.go
```diff
@@ -515,6 +512,34 @@ func mergeWithWorkflow[T any](value T, workflowValue *T) T {
...
+func getTempestRun(instance *testv1beta1.Tempest, workflowStepNum int) testv1beta1.TempestRunSpec {
+	if workflowStepNum < len(instance.Spec.Workflow) {
+		wtRun := instance.Spec.Workflow[workflowStepNum].TempestRun
+		tRun := changeType(wtRun)
+		return tRun
+	}
+	return instance.Spec.TempestRun
+}
+
+func changeType(wtRun testv1beta1.WorkflowTempestRunSpec) testv1beta1.TempestRunSpec {
```
lpiwowar, line 524: I think we should aim at making this function more generic by replacing the types with `interface{}`. The caller of the function will be responsible of casting the result to a correct type.

```suggestion
func changeType(instance interface{}) instance interface{} {
```

controllers/tempest_controller.go
```diff
@@ -515,6 +512,34 @@ func mergeWithWorkflow[T any](value T, workflowValue *T) T {
 	return *workflowValue
 }
 
+func getTempestRun(instance *testv1beta1.Tempest, workflowStepNum int) testv1beta1.TempestRunSpec {
```
lpiwowar, line 515: All the newly introduced functions should be moved to `common.go`.

[1] https://github.com/openstack-k8s-operators/test-operator/blob/main/controllers/common.go

lpiwowar (commented):
@kstrenkova, this is really nice progress! :tada: I say we schedule a call and discuss details but generally speaking I think we can make this work. I've just tried to test this locally. I had to make some minor tweaks (see comments) for the merging to work on the `TempestRun` section on my machine.

Anyway, really good job! :+1: Even when you do not manage to finish this task and somebody will be working on it instead of you then you still laid a really good foundation here! 

From our last conversation I was under the impression that this won't work. I think this will still require 2 - 3 PR updates and reviews but it is on a good path.

api/v1beta1/tempest_webhook.go
```diff
@@ -66,10 +67,55 @@ func (spec *TempestSpec) Default() {
...
+	value = mergeWorkflow(tcRun.Profile, wtcRun.Profile)
+	wtcRun.Profile = &value
+}
+
+// Return the right config values
+func mergeWorkflow[T any](value T, workflowValue *T) T {
+	if workflowValue == nil {
+		return value
+	}
+	return *workflowValue
```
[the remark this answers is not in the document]
kstrenkova: Deleted the example of non-iterative solution :+1:

api/v1beta1/tempest_webhook.go
```diff
@@ -66,10 +67,55 @@ func (spec *TempestSpec) Default() {
...
+		for key, _ := range spec.Workflow {
+			spec.mergeTempestRunIntoWorkflow(key)
+		}
+	}
+}
+
+// TODO Example of iterative solution
+func (spec *TempestSpec) mergeTempestRunIntoWorkflow(workflowStepNum int) {
+	tRun := spec.TempestRun
+	wtRun := spec.Workflow[workflowStepNum].TempestRun
```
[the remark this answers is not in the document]
kstrenkova: Oh, thank you, makes sense.

api/v1beta1/tempest_webhook.go
```diff
@@ -66,10 +67,55 @@ func (spec *TempestSpec) Default() {
...
+		tRunName := tRunReflect.Type().Field(i).Name
+		tRunValue := tRunReflect.Field(i)
+
+		wRunValue := wtRunReflect.FieldByName(tRunName)
+		if wRunValue.IsZero() {
+			tRunPtr := reflect.New(tRunValue.Type())
+			tRunPtr.Elem().Set(tRunValue)
+			wRunValue.Set(tRunPtr)
+		}
+	}
```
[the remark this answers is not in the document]
kstrenkova: Didn't work on this yet, but I can see what you mean. Let's discuss details during call.

api/v1beta1/tempest_webhook.go
```diff
@@ -66,10 +67,55 @@ func (spec *TempestSpec) Default() {
...
 	}
+	if len(spec.Workflow) > 0 {
+		for key, _ := range spec.Workflow {
+			spec.mergeTempestRunIntoWorkflow(key)
+		}
+	}
+}
+
+// TODO Example of iterative solution
+func (spec *TempestSpec) mergeTempestRunIntoWorkflow(workflowStepNum int) {
```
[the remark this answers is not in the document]
kstrenkova: I tried moving it in the next PR, I will push it soon.

kstrenkova:
I have added a new version of this patch that can also work with structures by calling the function for adding values from TempestRun section recursively. There is maybe needed a bit more generalization to get the functions working for all resources, but for Tempest this version should work.

lpiwowar (changes requested):
I think the PR is moving in the right direction. Nice work! :+1: 

I've added couple of comments @kstrenkova. Let me know what you think:).

api/v1beta1/common_webhook.go
```diff
@@ -28,3 +33,42 @@ const (
...
+	setNonZeroValues(tRunReflect, wtRunReflect, false)
+}
+
+func setNonZeroValues(src reflect.Value, dest reflect.Value, is_struct bool) {
+        for i := 0; i < src.NumField(); i++ {
+                tRunName := src.Type().Field(i).Name
+                tRunValue := src.Field(i)
+                wtRunValue := dest.FieldByName(tRunName)
+
+                if wtRunValue.IsZero() && !tRunValue.IsZero() {
```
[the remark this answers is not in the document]
kstrenkova: Nice catch, I think the need for `!tRunValue.IsNil` is still there though :thinking: The problem was that if the value in tempestRun was nil, it would create an error.

api/v1beta1/common_webhook.go
```diff
@@ -28,3 +33,42 @@ const (
 		"ensures that the copying of the logs to the PV is completed without any " +
 		"complications."
 )
+
+// merge non-workflow section into workflow
+func mergeSectionIntoWorkflow(instance interface{}, workflowStepNum int) {
```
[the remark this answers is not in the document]
kstrenkova: I looked into it and I don't understand why this would be needed :smile: Could you please explain your thought behind it a bit more?

api/v1beta1/common_webhook.go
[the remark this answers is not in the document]
lpiwowar: Yes, you are right:

```
The Tempest "tempest-tests" is invalid:
* spec.workflow[0].tempestRun.extraImagesType: Invalid value: "null": spec.workflow[0].tempestRun.extraImagesType in body must be of type array: "null"
* spec.workflow[0].tempestRun.externalPlugin: Invalid value: "null": spec.workflow[0].tempestRun.externalPlugin in body must be of type array: "null"
* spec.workflow[0].tempestRun.extraRPMs: Invalid value: "null": spec.workflow[0].tempestRun.extraRPMs in body must be of type array: "null"
```

Notice that the error only impacts arrays. We are trying to assign null to what should be a pointer to an array. Is there a way how to fix this? What about checking whether the array is empty in the tRun section and if it is then do not perform the assignment?

controllers/tobiko_controller.go
```diff
@@ -387,50 +388,46 @@ func (r *TobikoReconciler) PrepareTobikoEnvVars(
...
+		if instance.Spec.Workflow[workflowStepNum].SELinuxLevel != nil {
+			instance.Spec.SELinuxLevel = *instance.Spec.Workflow[workflowStepNum].SELinuxLevel
 		}
 
-		if instance.Spec.Workflow[step].Resources != nil {
-			instance.Spec.Resources = *instance.Spec.Workflow[step].Resources
+		if instance.Spec.Workflow[workflowStepNum].Resources != nil {
+			instance.Spec.Resources = *instance.Spec.Workflow[workflowStepNum].Resources
 		}
 	}
```
lpiwowar, line 413: **question:** Can we remove this code?

If the implementation works correctly then in the end this code should be removed altogether. 

**nit (non-blocking)**: Also, try to avoid renaming of variables unless absolutely necessary when doing bigger PRs or at least make it a separate commit. This makes the review a little bit more complicated as renaming of `step` to `workflowStepNum` is not 100 % connected to this PR.

controllers/common.go
```diff
@@ -693,3 +693,62 @@ func EnsureCloudsConfigMapExists(
...
+			workflow := typedInstance.Spec.Workflow[workflowStepNum]
+			mergeSections(&typedInstance.Spec, workflow)
+		}
+		return typedInstance.Spec
+	default:
+		return nil
+	}
+}
+
+func mergeSections(main interface{}, workflow interface{}) {
```
lpiwowar, line 726: **question:** Can you remind me. Why do we need this function when we are performing the merging into the workflow in the hook?

api/v1beta1/common_webhook.go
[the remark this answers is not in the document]
lpiwowar: **question:** Just curious. The modifications I suggested above were not possible?

lpiwowar (commented):
Just a couple of questions:). Overall moving in a good direction.

controllers/tobiko_controller.go
kstrenkova: I am planning on removing this code, but first I want the patch to work. I also do not know how this works precisely, so I left it there for now.

I renamed the variable because everywhere it was `workflowStepNum` except here. But I can split the commits when it's done, or if it's not too difficult then even now.

controllers/common.go
kstrenkova: So in my understanding, if we want to get for example `tempestRun` and there is no workflow we take value from the main section. If there is a workflow section, we can't just take from the workflow spec because the Types are different there than for main section. If we want it to work in both scenarios, we need to set it like main section, so it gets correct types.

But I also don't like the duplicate logic in this. Is there a way to make the types more compatible than doing it this way :thinking:

api/v1beta1/common_webhook.go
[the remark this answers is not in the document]
kstrenkova: I tried it, but it didn't help much with having to differentiate the type. I wanted the code to work first and then take care of details.

controllers/common.go
lpiwowar: Ok, now I remember. Do you think we could maybe make the types the same both in the main section and in the workflow section? This way we could theoretically get rid of the function and it could simplify some things. But it would require at least testing it out:). I'm leaving it up to you.

controllers/tobiko_controller.go
lpiwowar: Ok, good:). I'm keeping this comment open so that we do not forget about this.

Regarding the second point. Try to think about a PR as story. You're trying to make it as easy as possible for the reviewer. By splitting the PR into multiple commits it is much more easier to see the chunks of code that are coupled together (each commit is like a chapter). But no rush with this. It just caught my attention when going through the code.

controllers/common.go
kstrenkova: It is a possibility, but I am not sure if that's not too big of a change :thinking: If you remember, why is the Workflow not the same but pointers? And sometimes not even pointers.. Maybe there is a reason why they should be pointers and then the function is a good idea.

The function will not be too difficult to have, if Workflow section structures stay as they are, but I do agree that it would look nicer without it. Less work for controller as well, everything would be in webhook. So all in all, I am down for both options. However, I am not sure what impact would changing types have. If it was not a problem, I would assume it would be done like that already.

lpiwowar: > It is a possibility, but I am not sure if that's not too big of a change 🤔 If you remember, why is the Workflow not the same but pointers? 

The reason why we use pointers in the workflow section is because we need to be sure whether a value in the workflow section was set by a user or not. If the value is not set by the user then it is simply a `nil`. If we do not use pointers than the the values are set to a default value e.g for string `""`. 

It might be a big change but if it works then I believe in this case it would be justified.

Again, I'm leaving this up to your judgment. But I believe if we can get rid of the function it would be better.

kstrenkova: That's tricky. It sounds like it has a good reason to be as pointers then. But maybe the justification can be done :thinking: I also want to get rid of the function, but don't want to change how things are done too much.. As a person who was in charge of test-operator, would you say removing the pointer would be possible without big repercussions? If yes, I can go with the new strategy and push the changes for you to review.

lpiwowar: It is not that tricky as it might appear at first glance. As a person who was in charge of test-operator my answer is that it is at least worth trying. All you have to do is to just:

- set the types in the main section to pointers ([here](https://github.com/openstack-k8s-operators/test-operator/blob/main/api/v1beta1/tempest_types.go))
- If you are using VS Code for example it is going to scream at you at all the places where the types are mismatched. You fix those issues.
- You run test-operator and investigate whether it works or not.


>  If yes, I can go with the new strategy and push the changes for you to review.

Yes, but first please check that the code works as expected on your local environment.

>  I also want to get rid of the function, but don't want to change how things are done too much..

The main point of this PR is to simplify the code. Do not be afraid of changing the stuff. It is ok:)

api/v1beta1/common_webhook.go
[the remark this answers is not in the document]
kstrenkova: Finally got around to making this work. It was more simple than I thought.

controllers/common.go
kstrenkova: I didn't expect you to suggest setting the main section to pointers. I though the change would be in the workflow section. Then it would be difficult to get when a parameter is empty :smile: Which was my concern, but you managed to reply earlier.

I think it would be possible to take the route of pointers, but I don't think it would look that great and it is changing code on a lot of places. I think the current function does a lot of code simplifying already. The duplicate code among CR types will be gone and it's only needed to call the function once :thinking: Should I still go for the `types.go` changes? I am still unsure.

Yes, I will check that the code works for the final review request :D I am just trying to get the structure with you right now.

lpiwowar: > Should I still go for the types.go changes? 

As stated in my previous comment I believe it is worth trying. We can talk about it at the office on Thursday and updated the thread here after:).

>  I don't think it would look that great 

Why?

kstrenkova: Yes, let's talk it through in the office. The reason why I think it won't look that great is the difference it will create between operators. It is also true that workflow isn't present in them, but if we add a feature, I think the feature should be accommodating to the main functionality. Though, those can be considered minor inconveniences. Let's talk and find the best solution :smile:

kstrenkova:
recheck

kstrenkova:
I am gradually testing the most parameters I can. I will edit this comment when I test more parameters.
For now I have tested and they work:
- TempestRun: includeList, excludeList
- TempestconfRun: insecure

kstrenkova:
recheck

kstrenkova:
recheck

kstrenkova:
recheck

.zuul.yaml
```diff
@@ -11,6 +11,8 @@
               - openstack-k8s-operators-content-provider
             vars:
               cifmw_test_operator_bundle: "{{ content_provider_registry_ip }}:5001/openstack-k8s-operators/test-operator:{{ zuul.patchset }}"
+              cifmw_test_operator_repo: https://github.com/openstack-k8s-operators/test-operator.git
+              cifmw_test_operator_version: "{{ zuul.patchset }}"
```
kstrenkova, line 15: These variables can be erased for now, so we can merge the main changes introduced here. They were used for testing, but we can add them later when the change is merged in ci-framework.

lpiwowar (commented):
Let's go through this together on Wednesday:). Right now, I do not see any reason why this could not be merged :tada:

PR #170, opened 2024-08-14
Fix test-operator being stuck on NetworkAttachments error

lpiwowar:
This PR [1] introduced a bug that causes test-operator to be stuck on validation of the netowrkAttachments. This is caused by moving the code before the section that is responsible for the job creation. Basically, we are validating whether networkAttachments are correctly configured when there is no test pod to be checked.

This PR ensures that we are raising an error only when a job was created and the networkAttachments are not working.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/135

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -351,10 +351,12 @@ func (r *Reconciler) AcquireLock(
 		return err == nil, err
 	}
 
-	return false, nil
+	return false, err
 }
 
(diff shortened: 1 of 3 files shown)


lpiwowar:
ci/prow/test-operator-build-deploy

kopecmartin (approved):
thanks!

PR #173, opened 2024-08-15
Enable RuntimeDefault SeccompProfile 

lpiwowar:
It's a good practice to run containers with RuntimeDefault SeccomProfile that allows only such a system calls that can be expected to be used during normal operation of the container.


lpiwowar:
@kopecmartin can you add lgtm please if you agree with this PR?

kopecmartin (approved):
lgtm

PR #174, opened 2024-08-15
Enable networkAttachments testing in the check job

lpiwowar:


lpiwowar:
@kopecmartin I think this is ready to be merged.

SeanMooney (commented):
ack thanks for following up with the test coverage for this.

this looks reasonable to me to start with.

kopecmartin (approved):
thank you

PR #187, opened 2024-08-28
[18.0.0-proposed] Fix test-operator being stuck on NetworkAttachments error

kopecmartin:
This PR [1] introduced a bug that causes test-operator to be stuck on validation of the netowrkAttachments. This is caused by moving the code before the section that is responsible for the job creation. Basically, we are validating whether networkAttachments are correctly configured when there is no test pod to be checked.

This PR ensures that we are raising an error only when a job was created and the networkAttachments are not working.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/135

This is a cherry-pick of https://github.com/openstack-k8s-operators/test-operator/pull/170

### DIFF
--- a/controllers/tobiko_controller.go
+++ b/controllers/tobiko_controller.go
@@ -176,6 +176,29 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, fmt.Errorf("failed create network annotation from %s: %w",
 			instance.Spec.NetworkAttachments, err)
 	}
+
+	// NetworkAttachments
+	networkReady, networkAttachmentStatus, err := nad.VerifyNetworkStatusFromAnnotation(ctx, helper, instance.Spec.NetworkAttachments, serviceLabels, 1)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


controllers/tobiko_controller.go
```diff
@@ -176,6 +176,29 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			condition.NetworkAttachmentsReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.NetworkAttachmentsReadyErrorMessage,
+			err.Error()))
+
+		return ctrl.Result{}, err
+	}
+	// NetworkAttachments - end
+
```
lpiwowar, line 201: The code is twice here (line 262) and also the job is failing because of:

```
controllers/tempest_controller.go:360:45: no new variables on left side of :=
controllers/tobiko_controller.go:279:45: no new variables on left side of :=
```

lpiwowar (commented):
I'm going to take a look on this.

lpiwowar (approved):
LGTM

PR #191, opened 2024-09-02
Allow tcib depends on

lpiwowar:
Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/201


lpiwowar:
recheck

lpiwowar:
I'm not going to work on this anymore and this approach does not seem to be working ATM.

PR #192, opened 2024-09-02
Kubebuilder validation and doc style

kopecmartin:
The commit fixes a few doc issues, adds missing kubebuilder validations and unifies the style of the default kubebuilder validation.


lpiwowar:
Just FYI there seems to be some issue with zuul. The jobs are hanging for 15 and more hours for other projects as well.

lpiwowar:
/retest

Something is wrong with prow.

lpiwowar:
It looks like you have installed a different version of kubebuilder. See the difference between my `test.openstack.org_ansibletests.yaml` and yours:
```
           apiVersion:
-            description: |-
-              APIVersion defines the versioned schema of this representation of an object.
-              Servers should convert recognized schemas to the latest internal value, and
-              may reject unrecognized values.
-              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string

```
That's the reason why the precommit checks are failing. Just out of curiosity if you run from within the test-operator dir the following command then what is the output?
```
./bin/controller-gen --version
```

We can resolve this tomorrow.

kopecmartin:
I had to bump version of CONTROLLER_TOOLS_VERSION from 0.11.1 to 0.14.0 to workaround the segmentation fault i had:
https://github.com/openstack-k8s-operators/test-operator/blob/main/Makefile#L201C1-L201C25

lpiwowar:
On my env I use `0.11.1`. I think this is the version that is used in pre-commit checks as well. We should use this version. I can lend you my env if you want.

api/v1beta1/common.go
```diff
@@ -19,6 +19,7 @@ package v1beta1
 type extraConfigmapsMounts struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength:=253
```
lpiwowar, line 22: Is this something we want to add to other name parameters as well (e.g., `stepName`). Why 253? I'm just curious.

api/v1beta1/common.go
```diff
@@ -29,7 +30,7 @@ type extraConfigmapsMounts struct {
 
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:validation:optional
-	// +kubebuilder:default=""
+	// +kubebuilder:default:="/"
```
lpiwowar, line 33: This should stay `""`. More info here:

```
// Path within the volume from which the container's volume should be mounted.
// Defaults to "" (volume's root).
// +optional
SubPath [string](https://pkg.go.dev/builtin#string) `json:"subPath,omitempty" protobuf:"bytes,4,opt,name=subPath"`
```

[1] https://pkg.go.dev/k8s.io/api/core/v1#VolumeMount

api/v1beta1/ansibletest_types.go
```diff
@@ -29,73 +29,73 @@ type AnsibleTestSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:validation:Optional
 	// Extra configmaps for mounting in the pod.
-	ExtraMounts []extraConfigmapsMounts `json:"extraMounts,omitempty"`
+	ExtraConfigmapsMounts []extraConfigmapsMounts `json:"ExtraConfigmapsMounts,omitempty"`
```
lpiwowar, line 32: Sorry, for not being to explicit. 

```
ExtraConfigmapsMounts []extraConfigmapsMounts `json:"extraConfigmapsMounts,omitempty"`
```

(take a look at the other parameters)

`json:` <- this is name for the parameter in Tempest CR
`ExtraConfigmapsMounts` <- this is name of the parameter in the test-operator code

api/v1beta1/common.go
kopecmartin: no clue why 253, i copied it from the global operator .. i don't know if we want to add it to other name parameters, do we? I didn't notice other operators doing this

api/v1beta1/tempest_types_workflow.go
```diff
@@ -277,10 +277,12 @@ type WorkflowTempestSpec struct {
...
 	NetworkAttachments *[]string `json:"networkAttachments,omitempty"`
 
-	// BackoffLimimt allows to define the maximum number of retried executions (defaults to 6).
+	// BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
+	// +kubebuilder:validation:Optional
+        // +kubebuilder:default:=0
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:number"}
-	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
+	BackoffLimit *int32 `json:"backoffLimit"`
```
lpiwowar, line 285: Parameters in workflow section should not have default value. This section should stay the same. Except for:
```
// BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
```

api/v1beta1/tobiko_types.go
```diff
@@ -220,12 +221,15 @@ type TobikoWorkflowSpec struct {
         // Container image for tobiko
         ContainerImage string `json:"containerImage,omitempty"`
 
-        // BackoffLimimt allows to define the maximum number of retried executions (defaults to 6).
+        // BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:default:=0
         // +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:number"}
-        BackoffLimit *int32 `json:"backoffLimit,omitempty"`
+        BackoffLimit *int32 `json:"backoffLimit"`
```
lpiwowar, line 228: Parameters in workflow section should not have default value. This section should stay the same. Except for:
```
// BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
```

lpiwowar (commented):
I went through this PR and I have couple of comments. Once those gets fixed and I  check that kubeconfigSecretSecret works as expected we can probably merge this.

api/v1beta1/ansibletest_types.go
```diff
@@ -205,7 +208,8 @@ type AnsibleTestWorkflowSpec struct {
 	ContainerImage string `json:"containerImage,omitempty"`
 
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
-	// BackoffLimimt allows to define the maximum number of retried executions (defaults to 6).
+	// BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
+	// +kubebuilder:default:=0
```
lpiwowar, line 212: AnsibleTest workflow section should not have default value (the same way as Tobiko and Tempest)
```suggestion
	// +kubebuilder:validation:Optional
```

api/v1beta1/ansibletest_types.go
```diff
@@ -128,10 +129,12 @@ type AnsibleTestWorkflowSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:validation:Optional
 	// Extra configmaps for mounting in the pod
-	ExtraMounts []extraConfigmapsMounts `json:"extraMounts,omitempty"`
+	ExtraConfigmapsMounts []extraConfigmapsMounts `json:"extraConfigmapsMounts,omitempty"`
 
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength:=100
+	// +kubebuilder:default:=""
```
lpiwowar, line 137: Let's keep the default value unset. I know tobiko has the default value set for stepName. Ella is working on unifying all the common parameters and this will be one of the requirements. 
```suggestion
```

lpiwowar (commented):
Just two small changes needed.

lpiwowar:
Looks good to me, I think that we should check whether there are any jobs that consume the test-operator directly from upstream and that consume the `AnsibleTests.Spec.extraConfigmapsMounts` parameter. If not then I think that we can continue with:

1. Merging the test-operator change (I'm going to do that on Monday. I want to be here if something stops working for Ella).
2. Merging the ci-framework change.
3. Changing downstream jobs so that it sets both `cifmw_test_operator_extra_configmaps_mounts` and `cifmw_test_operator_extra_mounts`. Once the change reaches downstream we can remove `cifmw_test_operator_extra_mounts`.

PR #203, opened 2024-09-17
Override dhcp_domain flag for FQDN test

ashu-011:
test_verify_hostname_allows_fqdn test is failing downstream . This patch overrides dhcp_domain to '' as hostname in metadata doesn't have any suffix for it downstream


lpiwowar (approved):
LGTM :100: 

Thank you for the help! :+1:

lpiwowar:
> test_verify_hostname_allows_fqdn test is failing downstream 

@ashu-011, This is not a downstream job but I get what you mean:)

PR #212, opened 2024-09-27
Fix mistake in horizon_controller.go

ashu-011:
This commit fixes mistake in GetLogger function in horizon_controller.go as to make it correctly work for horizontest


lpiwowar (approved):
Thank you @ashu-011!

PR #218, opened 2024-10-04
Unify webhook error messages

lpiwowar:
This patch unifies the error messages returned by the test operator
webhooks. Previously, some error messages were returned from the
validateCreate as a plain errors. This made the error messages
unreadable.
    
This patch returns all the errors in field.ErrorList. Each error is
properly categorized with Type and Detail err message.


lpiwowar:
@kstrenkova, can you please do a review when you have time?:)

kopecmartin (approved):
lgtm

PR #205, opened 2024-09-20
Limit resources

lpiwowar:
This patch makes sure that there is a resource limit for each pod spawned by the test-operator. The limit is set high enough to ensure proper functioning of the test-operator but low enough to prevent resource draining in the ocp cluster:

- CPU limit: 2000m
- Memory limit: 4Gi

PR #222, opened 2024-10-08
Increase ResourceMemory limit

lpiwowar:
We are hitting an issue when pods are being killed in jobs because of OOM error. It seems like the ResourceMemory limit set with this PR [1] is not enough for the test operator pods.

This patch increases the ResourceMemory limit to 8Gi. We should revisit this problem later and investigate whether the limit can not be lower and what exactly caused the OOM errors.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/205


karelyatin:
This is causing issues in CI https://github.com/openstack-k8s-operators/edpm-ansible/pull/777 , this also increased Request along with limit making pod not to schedule at all.

PR #223, opened 2024-10-09
Update sample CRs in docs

ashu-011:


config/samples/test_v1beta1_horizontest.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
 spec:
   containerImage: ""
   # debug: false
+  storageClass: "lvms-local-storage"
```
lpiwowar, line 14: We should put here the default value which is `local-storage`. 
```suggestion
  storageClass: "local-storage"
```

lpiwowar (changes requested):

PR #224, opened 2024-10-09
Remove resource limit

lpiwowar:
This patch removes the resource limit for test operator pods as we are hitting both issue when:

- the limit is too low
- the limit is too high (scheduler does not have where to schedule the pod)

This is a hotfix before we expose the resource limit values through the test-operator CRs and find the correct default values.

Related PRs:
- https://github.com/openstack-k8s-operators/test-operator/pull/222
- https://github.com/openstack-k8s-operators/test-operator/pull/205


gibizer (approved):
Look OK to me as a hotfix.

kopecmartin (approved):

PR #225, opened 2024-10-11
Fix lint issues

lpiwowar:
This patch addresses two issues:

- incorrect indentation in common_webhook.go
- removal of automatically generated comments


lpiwowar:
@kopecmartin this can be merged if the job passes and you agree with the change. I split up [this](https://github.com/openstack-k8s-operators/test-operator/pull/211). It was getting chunkier and chunkier. I think on Monday we can merge https://github.com/openstack-k8s-operators/test-operator/pull/211 as well.

PR #226, opened 2024-10-11
Add build index script 

lpiwowar:
Add a script that can be used to quickly build index containing
test operator. The script pushes the index to quay.io by default.

Example of useage:

USER=quayio-nick TAG=xyz ./hack/build_index.sh

### DIFF
--- a/hack/build_index.sh
+++ b/hack/build_index.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# This command builds test-operator index and pushes it to quay.io. Podman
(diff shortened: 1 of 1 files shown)


lpiwowar:
@kopecmartin this can be merged if the job passes and you agree with the change. I split up this -> [https://github.com/openstack-k8s-operators/test-operator/pull/211]. It was getting chunkier and chunkier. I think on Monday we can merge #211 as well.

kopecmartin (approved):
lgtm

PR #231, opened 2024-10-18
[OSPRH-10386] Update logic for workflows

lpiwowar:
There was a race condition in the previous implementation of the
workflows that was caused by the dependency of the workflow feature
on a ConfigMap that served as counter.

This commit simplifies the logic of the Reconcile loop in two ways:

  1. Removes the dependency for an external workflow counter

  2. Introduces NextAction function which decides what next action
     should be performed by the Reconcile loop based on the current
     state of the OpenShift cluster. The input of this function is the
     instance which is currently being processed and a workflowLength.
     Using these two arguments it then tells the Reconcile loop which
     actions it should perform: Wait, CreateFirstJob, CreateNextJob,
     EndTesting, Failure.

This approach should simplify the reconcile logic and move the
test-operator towards a unified Reconcile loop.

### DIFF
--- a/controllers/ansibletest_controller.go
+++ b/controllers/ansibletest_controller.go
@@ -18,6 +18,8 @@ package controllers
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"strconv"
 	"time"
 
@@ -37,7 +39,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -68,9 +69,6 @@ func (r *AnsibleTestReconciler) GetLogger(ctx context.Context) logr.Logger {
 func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
 	Log := r.GetLogger(ctx)
 
-	// How much time should we wait before calling Reconcile loop when there is a failure
-	requeueAfter := time.Second * 60
-
 	// Fetch the ansible instance
 	instance := &testv1beta1.AnsibleTest{}
 	err := r.Client.Get(ctx, req.NamespacedName, instance)
@@ -79,11 +77,6 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, err
-	}
-
-	workflowActive := false
-	if len(instance.Spec.Workflow) > 0 {
-		workflowActive = true
 	}
 
 	// Create a helper
@@ -142,40 +135,62 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	}
 
-	// Ensure that there is an external counter and read its value
-	// We use the external counter to keep track of the workflow steps
-	r.WorkflowStepCounterCreate(ctx, instance, helper)
-	externalWorkflowCounter := r.WorkflowStepCounterRead(ctx, instance)
-	if externalWorkflowCounter == -1 {
-		return ctrl.Result{RequeueAfter: requeueAfter}, nil
-	}
+	workflowLength := len(instance.Spec.Workflow)
+	nextAction, nextWorkflowStep, err := r.NextAction(ctx, instance, workflowLength)
 
-	// Each job that is being executed by the test operator has
-	currentWorkflowStep := 0
-	runningAnsibleJob := &batchv1.Job{}
-	runningJobName := r.GetJobName(instance, externalWorkflowCounter-1)
-	err = r.Client.Get(ctx, client.ObjectKey{Namespace: instance.GetNamespace(), Name: runningJobName}, runningAnsibleJob)
-	if err != nil && !k8s_errors.IsNotFound(err) {
+	switch nextAction {
+	case Failure:
 		return ctrl.Result{}, err
-	} else if err == nil {
-		currentWorkflowStep, _ = strconv.Atoi(runningAnsibleJob.Labels["workflowStep"])
-	}
 
-	if r.CompletedJobExists(ctx, instance, currentWorkflowStep) {
-		// The job created by the instance was completed. Release the lock
-		// so that other instances can spawn a job.
-		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
-		Log.Info("Job completed")
+	case Wait:
+		Log.Info(InfoWaitingOnJob)
+		return ctrl.Result{RequeueAfter: RequeueAfterValue}, nil
+
+	case EndTesting:
+		// All jobs created by the instance were completed. Release the lock
+		// so that other instances can spawn their jobs.
 		if lockReleased, err := r.ReleaseLock(ctx, instance); !lockReleased {
-			return ctrl.Result{}, err
+			Log.Info(fmt.Sprintf(InfoCanNotReleaseLock, testOperatorLockName))
+			return ctrl.Result{RequeueAfter: RequeueAfterValue}, err
 		}
+
+		instance.Status.Conditions.MarkTrue(
+			condition.DeploymentReadyCondition,
+			condition.DeploymentReadyMessage)
+
+		Log.Info(InfoTestingCompleted)
+		return ctrl.Result{}, nil
+
+	case CreateFirstJob:
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, false)
+		if !lockAcquired {
+			Log.Info(fmt.Sprintf(InfoCanNotAcquireLock, testOperatorLockName))
+			return ctrl.Result{RequeueAfter: RequeueAfterValue}, err
+		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingFirstPod, nextWorkflowStep))
+
+	case CreateNextJob:
+		// Confirm that we still hold the lock. This is useful to check if for
+		// example somebody / something deleted the lock and it got claimed by
+		// another instance. This is considered to be an error state.
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, false)
+		if !lockAcquired {
+			Log.Error(err, ErrConfirmLockOwnership, testOperatorLockName)
+			return ctrl.Result{RequeueAfter: RequeueAfterValue}, err
+		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingNextPod, nextWorkflowStep))
+
+	default:
+		return ctrl.Result{}, errors.New(ErrReceivedUnexpectedAction)
 	}
 
 	serviceLabels := map[string]string{
 		common.AppSelector: ansibletest.ServiceName,
-		"workflowStep":     strconv.Itoa(externalWorkflowCounter),
-		"instanceName":     instance.Name,
-		"operator":         "test-operator",
+		workflowStepLabel:  strconv.Itoa(nextWorkflowStep),
+		instanceNameLabel:  instance.Name,
+		operatorNameLabel:  "test-operator",
 	}
 
 	// Create PersistentVolumeClaim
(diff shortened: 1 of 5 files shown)


controllers/horizontest_controller.go
```diff
@@ -134,25 +132,55 @@ func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 
-	if r.CompletedJobExists(ctx, instance, 0) {
-		// The job created by the instance was completed. Release the lock
-		// so that other instances can spawn a job.
-		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
-		Log.Info("Job completed")
-		if lockReleased, err := r.ReleaseLock(ctx, instance); !lockReleased {
-			return ctrl.Result{}, err
+	workflowLength := 0
+	nextAction, nextWorkflowStep, err := r.NextAction(ctx, instance, workflowLength)
```
slawqo, line 136: workflowLenght will always be 0 here, is that expected?

controllers/tempest_controller.go
```diff
@@ -159,45 +151,56 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
...
+		return ctrl.Result{}, errors.New(ErrReceivedUnexpectedAction)
 	}
 
 	serviceLabels := map[string]string{
 		common.AppSelector: tempest.ServiceName,
-		"workflowStep":     strconv.Itoa(externalWorkflowCounter),
-		"instanceName":     instance.Name,
-		"operator":         "test-operator",
+		workflowStepLabel:  strconv.Itoa(nextWorkflowStep),
+		instanceNameLabel:  instance.Name,
```
evallesp, line 197: (blocking) question: Is not necessary "operator" here?

controllers/tobiko_controller.go
```diff
@@ -148,38 +140,51 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+			condition.DeploymentReadyMessage)
+
+		Log.Info(InfoTestingCompleted)
+		return ctrl.Result{}, nil
+
+	case CreateFirstJob:
+		lockAcquired, _ := r.AcquireLock(ctx, instance, helper, instance.Spec.Parallel)
+		if !lockAcquired {
+			Log.Info(fmt.Sprintf(InfoCanNotAcquireLock, testOperatorLockName))
+			return ctrl.Result{}, nil
```
evallesp, line 171: (blocking)  question: Would be good to return RequeueAfterValue here? If not, could you share some info? Thanks!

evallesp:
In general really good patch IMHO.
It's good to have a single pattern across the different controllers, and I see the point (Apart from fixing a race condition) as a good work of refactoring.

Some extra things:
* Comments in the switch clauses at ansibletest_controller are also applicable to other controllers.
* It seems there's room for some refactoring, as the code between controllers' logic of switch clauses is quite similar.

I'd like to do a second round when my comments are addressed, but it's IMHO in good shape and near to being merged. Thanks!

controllers/ansibletest_controller.go
```diff
@@ -142,40 +135,51 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			condition.DeploymentReadyMessage)
+
+		Log.Info(InfoTestingCompleted)
+		return ctrl.Result{}, nil
+
+	case CreateFirstJob:
+		lockAcquired, _ := r.AcquireLock(ctx, instance, helper, false)
+		if !lockAcquired {
+			Log.Info(fmt.Sprintf(InfoCanNotAcquireLock, testOperatorLockName))
+			return ctrl.Result{RequeueAfter: RequeueAfterValue}, nil
```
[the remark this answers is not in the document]
evallesp: Not updated here

controllers/horizontest_controller.go
```diff
@@ -134,25 +132,66 @@ func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		}
+
+		instance.Status.Conditions.MarkTrue(
+			condition.DeploymentReadyCondition,
+			condition.DeploymentReadyMessage)
+
+		Log.Info(InfoTestingCompleted)
+		return ctrl.Result{}, nil
+
+	case CreateFirstJob:
```
evallesp, line 161: (blocking) suggestion: err message here is not available. I guess we also need to add it here.

controllers/ansibletest_controller.go
```diff
@@ -142,40 +135,51 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingFirstPod, nextWorkflowStep))
+
+	case CreateNextJob:
```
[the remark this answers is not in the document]
evallesp: I'm OK if you ended up not adding that after your explanation.
I've noticed you did add it.

Solving this conversation.

controllers/common.go
```diff
@@ -52,6 +71,126 @@ type Reconciler struct {
 	Scheme  *runtime.Scheme
 }
 
+// NextAction holds an action that should be performed by the Reconcile loop.
+type NextAction int
+
+const (
+	// Wait indicates that we should wait for the state of the OpenShift cluster
+	// to change
+	Wait = iota
```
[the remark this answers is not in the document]
evallesp: Works for me, thanks for the explanation!

controllers/common.go
```diff
@@ -368,6 +508,10 @@ func (r *Reconciler) AcquireLock(
 		return err == nil, err
 	}
 
+	if cm.Data[testOperatorLockOnwerField] == instanceGUID {
```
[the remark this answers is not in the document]
evallesp: Ack, understood! Thanks.

controllers/tempest_controller.go
evallesp: ack!

evallesp (commented):
Found two places where it seems to be a missing err catch.

controllers/horizontest_controller.go
lpiwowar: Done, thanks :+1:

controllers/horizontest_controller.go
```diff
@@ -134,25 +132,66 @@ func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		Log.Info(fmt.Sprintf(InfoCreatingFirstPod, nextWorkflowStep))
+
+	case CreateNextJob:
+		// Confirm that we still hold the lock. This is useful to check if for
+		// example somebody / something deleted the lock and it got claimed by
+		// another instance. This is considered to be an error state.
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, instance.Spec.Parallel)
+		if !lockAcquired {
+			Log.Error(err, ErrConfirmLockOwnership, testOperatorLockName)
 			return ctrl.Result{}, err
```
evallesp, line 177: (blocking) question: IIRC, I left here a comment, but I don't see it now, Do we need here the requeueAfter here right?

controllers/horizontest_controller.go
```diff
@@ -134,25 +132,66 @@ func (r *HorizonTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingNextPod, nextWorkflowStep))
+
+	default:
+		return ctrl.Result{}, errors.New(ErrReceivedUnexpectedAction)
```
evallesp, line 183: (blocking) question: Do we need here the requeueAfter here right?

controllers/tempest_controller.go
```diff
@@ -159,45 +151,68 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
...
+		Log.Info(fmt.Sprintf(InfoCreatingFirstPod, nextWorkflowStep))
+
+	case CreateNextJob:
+		// Confirm that we still hold the lock. This is useful to check if for
+		// example somebody / something deleted the lock and it got claimed by
+		// another instance. This is considered to be an error state.
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, instance.Spec.Parallel)
+		if !lockAcquired {
+			Log.Error(err, ErrConfirmLockOwnership, testOperatorLockName)
 			return ctrl.Result{}, err
```
evallesp, line 196: (blocking) question: Do we need here the requeueAfter here right?

controllers/tobiko_controller.go
```diff
@@ -148,38 +140,62 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+			condition.DeploymentReadyMessage)
+
+		Log.Info(InfoTestingCompleted)
+		return ctrl.Result{}, nil
+
+	case CreateFirstJob:
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, instance.Spec.Parallel)
+		if !lockAcquired {
+			Log.Info(fmt.Sprintf(InfoCanNotAcquireLock, testOperatorLockName))
+			return ctrl.Result{}, err
```
evallesp, line 173: (blocking) question: Do we need here the requeueAfter here right?

controllers/tobiko_controller.go
```diff
@@ -148,38 +140,62 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+		Log.Info(fmt.Sprintf(InfoCreatingFirstPod, nextWorkflowStep))
+
+	case CreateNextJob:
+		// Confirm that we still hold the lock. This needs to be checked in order
+		// to prevent situation when somebody / something deleted the lock and it
+		// got claimedy by another instance.
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, instance.Spec.Parallel)
+		if !lockAcquired {
+			Log.Error(err, ErrConfirmLockOwnership, testOperatorLockName)
 			return ctrl.Result{}, err
```
evallesp, line 185: (blocking) question: Do we need here the requeueAfter here right?

controllers/tobiko_controller.go
```diff
@@ -148,38 +140,62 @@ func (r *TobikoReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
 		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingNextPod, nextWorkflowStep))
+
+	default:
+		return ctrl.Result{}, errors.New(ErrReceivedUnexpectedAction)
```
evallesp, line 191: (blocking) question: Do we need here the requeueAfter here right?

controllers/ansibletest_controller.go
```diff
@@ -142,40 +135,62 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		lockAcquired, err := r.AcquireLock(ctx, instance, helper, false)
+		if !lockAcquired {
+			Log.Error(err, ErrConfirmLockOwnership, testOperatorLockName)
 			return ctrl.Result{}, err
 		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingNextPod, nextWorkflowStep))
+
+	default:
+		return ctrl.Result{}, errors.New(ErrReceivedUnexpectedAction)
```
evallesp, line 186: (blocking) question: Do we need here the requeueAfter here right?

controllers/tempest_controller.go
```diff
@@ -159,45 +151,68 @@ func (r *TempestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
...
 		}
+
+		Log.Info(fmt.Sprintf(InfoCreatingNextPod, nextWorkflowStep))
+
+	default:
+		return ctrl.Result{}, errors.New(ErrReceivedUnexpectedAction)
```
evallesp, line 202: (blocking) question: Do we need here the requeueAfter here right?

evallesp (commented):
Added some other questions.

controllers/horizontest_controller.go
evallesp: Cool!

controllers/ansibletest_controller.go
```diff
@@ -142,40 +135,51 @@ func (r *AnsibleTestReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		if lockReleased, err := r.ReleaseLock(ctx, instance); !lockReleased {
-			return ctrl.Result{}, err
+	case Wait:
+		Log.Info(InfoWaitingOnJob)
+		return ctrl.Result{RequeueAfter: RequeueAfterValue}, nil
+
+	case EndTesting:
+		if lockReleased, _ := r.ReleaseLock(ctx, instance); !lockReleased {
+			Log.Info(fmt.Sprintf(InfoCanNotAcquireLock, testOperatorLockName))
+			return ctrl.Result{RequeueAfter: RequeueAfterValue}, nil
```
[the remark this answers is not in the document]
evallesp: Thanks

controllers/tobiko_controller.go
evallesp: Solving this comment, here it's not addressed the reququeEafter, but I see the err catching.
Solving this to not having two comments pointing the same issue.
My fault I didn't find this comment.

lpiwowar (commented):
Thanks @evallesp!:)

controllers/tobiko_controller.go
lpiwowar: I had a second though about this one. I think `RequeueAfter:` wouldn't hurt here. There might be an error when communicating with the OCP API.

controllers/tempest_controller.go
lpiwowar: I had a second though about this one. I think RequeueAfter: wouldn't hurt here. There might be an error when communicating with the OCP API.

controllers/horizontest_controller.go
lpiwowar: I had a second though about this one. I think RequeueAfter: wouldn't hurt here. There might be an error when communicating with the OCP API.

lpiwowar:
recheck

Failure not related to the test-operator:

```
 msg: 'Failed to download metadata for repo ''delorean-component-barbican'': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried'
```

lpiwowar:
recheck

PR #233, opened 2024-10-23
Introduce new parameter Expected Failures List

kstrenkova:
The Expected Failures List parameter will help with specifying set of tests whose failures we want to ignore. By adding test to this list, its failure will not have impact on the jobs pass/fail. It can help with not having to re-run long running jobs and save time and resources in specific cases, for example when dealing with flaky tests.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -57,6 +57,8 @@
         tempest.api.identity.admin.v3.test_credentials.CredentialsTestJSON
         tempest.api.identity.admin.v3.test_tokens.TokensV3TestJSON.test_rescope_token
         tempest.api.identity.admin.v3.test_users.UsersV3TestJSON.test_update_user_password
+      cifmw_test_operator_tempest_expected_failures_list: |
+        foobar
       cifmw_tempest_tempestconf_config:
         overrides: |
           compute-feature-enabled.dhcp_domain ''
--- a/api/v1beta1/tempest_types.go
+++ b/api/v1beta1/tempest_types.go
@@ -153,6 +153,13 @@ type TempestRunSpec struct {
 
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// The expectedFailuresList parameter contains tests that should not count
+	// as failures. When a test from this list fails, the test pod ends with
+	// Completed state rather than with Error state.
+	ExpectedFailuresList string `json:"expectedFailuresList"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// +kubebuilder:default:=0
 	// Concurrency value that is passed to tempest via --concurrency
 	Concurrency int64 `json:"concurrency"`
--- a/config/crd/bases/test.openstack.org_tempests.yaml
+++ b/config/crd/bases/test.openstack.org_tempests.yaml
@@ -168,6 +168,12 @@ spec:
                     description: A content of exclude.txt file that is passed to tempest
                       via --exclude-list
                     type: string
+                  expectedFailuresList:
+                    description: The expectedFailuresList parameter contains tests
+                      that should not count as failures. When a test from this list
+                      fails, the test pod ends with Completed state rather than with
+                      Error state.
+                    type: string
                   externalPlugin:
                     description: ExternalPlugin contains information about plugin
                       that should be installed within the tempest test pod. If this
@@ -602,6 +608,12 @@ spec:
                         excludeList:
                           description: A content of exclude.txt file that is passed
                             to tempest via --exclude-list
+                          type: string
+                        expectedFailuresList:
+                          description: The expectedFailuresList parameter contains
+                            tests that should not count as failures. When a test from
+                            this list fails, the test pod ends with Completed state
+                            rather than with Error state.
                           type: string
                         externalPlugin:
                           description: ExternalPlugin contains information about plugin
--- a/config/samples/test_v1beta1_tempest.yaml
+++ b/config/samples/test_v1beta1_tempest.yaml
@@ -46,6 +46,8 @@ spec:
     concurrency: 8
     # excludeList: | # <-- Use | to preserve \n
     #   tempest.api.identity.v3.*
+    # expectedFailuresList: | # <-- Use | to preserve \n
+    #   tempest.api.identity.v3.test_tokens.TokensV3Test.test_create_token
     # workerFile: | # <-- Use | to preserve \n
     #   - worker:
     #     - tempest.api.*
--- a/api/bases/test.openstack.org_tempests.yaml
+++ b/api/bases/test.openstack.org_tempests.yaml
@@ -168,6 +168,12 @@ spec:
                     description: A content of exclude.txt file that is passed to tempest
                       via --exclude-list
                     type: string
+                  expectedFailuresList:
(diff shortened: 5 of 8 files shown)


config/crd/bases/test.openstack.org_tempests.yaml
```diff
@@ -168,6 +168,10 @@ spec:
                     description: A content of exclude.txt file that is passed to tempest
                       via --exclude-list
                     type: string
+                  expectedFailuresList:
+                    description: A content of expected_failures.txt file that is passed
+                      to tempest via --expected-failures-list
```
lpiwowar, line 173: @kopecmartin do you think `--expected-failures-list` is something that would be "mergeable" in tempest.

kopecmartin: this doesn't belong to upstream tempest, tempest tests are meant to be explicit - https://opendev.org/openstack/tempest/src/branch/master/REVIEWING.rst#being-explicit
Parameter like this may mask failures (especially if a user forgets that a test result is being ignored) because they'll be ignored .. if tests fail due to a known bug, the user should not run them - if it's a bug in tempest, it can be skipped in tempest, if it's a bug in a component, well, the test did its job - if a user doesn't want a failing CI in the meantime, they can exclude such test.

that's an upstream perspective ^^

Downstream perspective:
I'm still not convinced this is really needed as any failing or potentially failing test can be simply excluded.

arxcruz: This have nothing to do with tempest in upstream. I expect to have a list of tests that I know it's unstable (timeout, concurrency issues, etc) and I want testoperator to check if the match list of expected failing tests matches the failing tests in a particular job, and if does, don't mark the job as failure. The tempest test will continue to fail, and that is okay, but in CI, the job will be marked as success and we have the failures reported to report portal/polarion. This is important to add unijobs into the promotion criteria, since unijobs run 5k tests, sometimes a test fails because of concurrency, or whitebox tests for example that are very unstable. But we want to collect the data of how many times it fails/pass so we can decide later what to do with these tests.

kopecmartin: all clear now, this in test-operator is gonna be an interface how to get the tests (that might fail) to tcib script where the "don't fail on these tests" logic will be written:
https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/tcib/base/os/tempest/run_tempest.sh

this won't reach the tempest upstream.

arxcruz: Exactly, this is a unijobs downstream feature only. Nothing related to upstream tempest.

lpiwowar: Ok, it seems we have an agreement and common understanding of how it should work. In this case just Katka, please, when you have time and this PR is ready, update the description of the `expectedFailuresList`. Right now it suggests that the `--expected-failures-list` parameter is passed to tempest.

kstrenkova: Thank you @arxcruz for making the intention for this parameter clear. It is my mistake for writing the description in such a way because I followed the code in include and exclude list. I fixed the description in the last push, so I hope from now on it will be smooth sailing. Sorry for the confusion :sweat_smile:

lpiwowar: That's ok, @kstrenkova. At least we sorted things out :)

lpiwowar (changes requested):
Just two small request otherwise looks good to me!:) :+1:

api/v1beta1/tempest_types.go
```diff
@@ -151,6 +151,12 @@ type TempestRunSpec struct {
 	// A content of exclude.txt file that is passed to tempest via --exclude-list
 	ExcludeList string `json:"excludeList"`
 
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// The expected_failures.txt file contains tests that should not count
+	// as failures for the job, as they are expected to fail.
```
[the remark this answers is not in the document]
kstrenkova: I agree with you :smile: I just did a minor fixes to the sentence.

config/samples/test_v1beta1_tempest.yaml
```diff
@@ -46,6 +46,8 @@ spec:
     concurrency: 8
     # excludeList: | # <-- Use | to preserve \n
     #   tempest.api.identity.v3.*
+    # expectedFailuresList: | # <-- Use | to preserve \n
+    #   tempest.api.identity.v3.*
```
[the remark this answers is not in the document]
kstrenkova: Okay, that makes sense. Thanks for pointing that out :+1:

kstrenkova:
I have make the requested changes and changed the definition of the `podified-multinode-edpm-deployment-crc-test-operator` job, so it uses newly introduced parameter.

.zuul.yaml
```diff
@@ -57,6 +57,7 @@
         tempest.api.identity.admin.v3.test_credentials.CredentialsTestJSON
         tempest.api.identity.admin.v3.test_tokens.TokensV3TestJSON.test_rescope_token
         tempest.api.identity.admin.v3.test_users.UsersV3TestJSON.test_update_user_password
+      cifmw_test_operator_tempest_expected_failures_list: |
```
lpiwowar, line 60:
```suggestion
      cifmw_test_operator_tempest_expected_failures_list: |
        foobar
```

lpiwowar (changes requested):
You can also add Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2513

We can test the ci-framework change together with this PR as well:).

lpiwowar:
> I have make the requested changes and changed the definition of the `podified-multinode-edpm-deployment-crc-test-operator` job, so it uses newly introduced parameter.

thanks! :tada:

.zuul.yaml
kstrenkova: Yeah, I was curious if the option can be set empty like this. Do I need to add foobar for it to work? :thinking:

lpiwowar: You could leave it empty but then it would be better to define the var like this (no need for `|` [1]).

```
cifmw_test_operator_tempest_expected_failures_list: ""
``` 
But I would prefer if you added the `foobar` value. Think about it in terms of code coverage. How much lines of the newly added code to test-operator and tcib would be executed if you lave the value stored in the variable empty in contrast with the `foobar` value? :thinking: [2]

It's definitely good to testy for edge cases like empty value but in this job it would be good if we tried to cover as much code as possible.

[1] [here](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics)
[2] [here](https://github.com/openstack-k8s-operators/test-operator/pull/233/files#diff-19fb6c8a471eb8f571e5e959d274704a582df8fa55a60f500593fcba5efb8823R448)

kstrenkova: Alright, thank you for your answer. That was an interesting view on this, I did not think about it in that way :smile:

lpiwowar (commented):
Looks good to me!:) :+1:  Let's wait for the tcib change to get merged first.

P.S.: If you want to improve a little bit the description of the parameter then I think it would be great to mention that the parameter does not accept regexes but only exact full test names.

lpiwowar:
recheck

lpiwowar:
recheck

PR #237, opened 2024-10-30
Limit pod resources

lpiwowar:
This patch introduces back limits for the pods spawned by the
test-operator after they were increased and later removed with these
two PRs [1][2].

The problem with the previous two patches was that they only set the
Resources.Limits field and not the Resources.Requests field. When
Resources.Limits is set and Resources.Requests is empty then it
inherrits the value from Resources.Limits.

Therefore, we first hit the OOM killed issue when we set the
Resources.Limits too low and later when we increased the value we hit
the "Insufficient memory" error (due to high value in
Resources.Requests field)

This patch addresses the above mentioned issue by:
  - setting sane default values for Resource.Limits
  - setting sane default values for Resource.Requests and
  - introduces new parameter called .Spec.Resources which can be used
    to change the default values.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/222
[2] https://github.com/openstack-k8s-operators/test-operator/pull/224


lpiwowar:
Moved here -> https://github.com/openstack-k8s-operators/test-operator/pull/253

PR #240, opened 2024-11-05
[Horizon][Tobiko] Do not fail when clouds CM exists

lpiwowar:
The Ensure[Horizon|Tobiko]CloudsYAML function was failing when the CM containing the modified clouds.yaml already existed:

  Object openstack/horizontest-clouds-config is already owned
  by another HorizonTest controller horizontest-tests

This patch does two things:

  - Ensures that we do not create the modified clouds.yaml when it already exists.

  - Removes the duplicate implementation of of the Ensure*CloudsYAML function and replaces it with EnsureCloudsConfigMapExists.


controllers/common.go
```diff
@@ -550,3 +551,71 @@ func GetCommonRbacRules(privileged bool) []rbacv1.PolicyRule {
...
+		helper,
+		instance,
+		openstackConfigMapName,
+		time.Second*10,
+	)
+	if cm.Name == testOperatorCloudsConfigMapName {
+		return ctrl.Result{}, nil
+	}
+
+	cm, _, _ = configmap.GetConfigMap(
```
kstrenkova, line 578: I am not sure why this part is present 2 times in the function. I don't know if it's a copy-paste mistake or is needed there.

lpiwowar (commented):
Thank you @kstrenkova! :pray:

lpiwowar:
recheck

kstrenkova (approved):
I have tested this change and it seems to be working. Lgtm.

controllers/common.go
```diff
@@ -550,3 +551,72 @@ func GetCommonRbacRules(privileged bool) []rbacv1.PolicyRule {
...
+// Some frameworks like (e.g., Tobiko and Horizon) require password value to be
+// present in clouds.yaml. This code ensures that we set a default value of
+// 12345678 when password value is missing in the clouds.yaml
+func EnsureCloudsConfigMapExists(
+	ctx context.Context,
+	instance client.Object,
+	helper *helper.Helper,
+	labels map[string]string,
+) (ctrl.Result, error) {
+	const openstackConfigMapName = "openstack-config"
```
evallesp, line 564: (blocking) question: I guess this one is not required right? Related as well with @kstrenkova 
comment.
Found TestOperatorCloudsConfigMapName in the volumes.go files.

controllers/common.go
```diff
@@ -550,3 +551,72 @@ func GetCommonRbacRules(privileged bool) []rbacv1.PolicyRule {
...
+	const testOperatorCloudsConfigMapName = "test-operator-clouds-config"
```
evallesp, line 565: (non-blocking) question: would take the name from pkg.util.common.go ? So we don't need to maintain same const in two different places.

evallesp (commented):
Some questions but in general lgtm.

controllers/common.go
evallesp: I don't see this fixed.

lpiwowar: The function should be called two times. The issue was that we called it with `openstackConfigMapName` in both function calls.

PR #243, opened 2024-11-13
[OSPRH-11191] Remove dereference for nil pointer

lpiwowar:
This patch fixes the issue when ExtraConfigmapsMounts is empty.
Dereferencing a nil pointer leads to a panic.

We now first check whether the pointer is nil and then move to
dereferencing it.


kopecmartin (approved):
lgtm

PR #247, opened 2024-11-18
[Workflows] Order job name alphabetically

eduolivares:
Usually directories are ordered alphabetically when we access to them via CLI or UI.
With this change the alphabetical order corresponds with the workflow order.


lpiwowar:
/approve

Looks good to me! Thank you @eduolivares! It looks much better this way.

PR #248, opened 2024-11-18
[18.0-fr1][OSPRH-11235] Do not fail when clouds CM exists

lpiwowar:
The Ensure[Horizon|Tobiko]CloudsYAML function was failing when the CM containing the modified clouds.yaml already existed:

  Object openstack/horizontest-clouds-config is already owned
  by another HorizonTest controller horizontest-tests

This patch does two things:

  - Ensures that we do not create the modified clouds.yaml when it already exists.

  - Removes the duplicate implementation of of the Ensure*CloudsYAML function and replaces it with EnsureCloudsConfigMapExists.


lpiwowar:
recheck

lpiwowar:
recheck

PR #250, opened 2024-11-22
Extend test-operator zuul job to enable horizon testing

ashu-011:
This patch extends test-operator zuul job to enable basic horizon testing which can be later extended to full testing . Currently the parent job used by test-operator zuul job does not setup cinder backend , this commit also changes the parent job to make use of another job which is already setups the cinder backend as ceph for the crc deployment .

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -22,7 +22,7 @@
 
 - job:
     name: podified-multinode-edpm-deployment-crc-test-operator
-    parent: podified-multinode-edpm-deployment-crc
+    parent: podified-multinode-hci-deployment-crc-1comp-backends
     vars:
       cifmw_install_yamls_whitelisted_vars:
         - 'TEST_REPO'
@@ -67,7 +67,7 @@
         tempest.api.identity.admin.v3.test_users.UsersV3TestJSON.test_update_user_password
       cifmw_test_operator_tempest_expected_failures_list: |
         foobar
-      cifmw_tempest_tempestconf_config:
+      cifmw_test_operator_tempest_tempestconf_config:
         overrides: |
           compute-feature-enabled.dhcp_domain ''
           identity.v3_endpoint_type public
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -41,6 +41,8 @@
           type: tobiko
         - name: ansibletest
           type: ansibletest
+        - name: horizontest
+          type: horizontest
 
       # Tempest
       cifmw_run_tempest: true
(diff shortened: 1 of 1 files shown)


lpiwowar:
@ashu-011, I just quickly skimmed through the job. It seems like there are not enough PVCs for the `horizontest` to be executed:
```
Events:
  Type     Reason            Age                   From               Message
  ----     ------            ----                  ----               -------
  Warning  FailedScheduling  3m10s (x63 over 63m)  default-scheduler  0/1 nodes are available: 1 node(s) didn't find available persistent volumes to bind. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
```

https://logserver.rdoproject.org/50/250/104ce9f5d113d674b538d40d635e375d40deef27/github-check/podified-multinode-edpm-deployment-crc-test-operator/6440b64/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pvc.log

ashu-011:
recheck

lpiwowar:
@ashu-011, I'm creating an autohold here -> https://github.com/openstack-k8s-operators/test-operator/pull/254. The job is failing with other PRs as well. It seems something is wrong with the job (not enough PVs?).

ashu-011:
> @ashu-011, I'm creating an autohold here -> #254. The job is failing with other PRs as well. It seems something is wrong with the job (not enough PVs?).

sure thanks @lpiwowar

ashu-011:
recheck

ashu-011:
recheck

ashu-011:
recheck

ashu-011:
recheck

ashu-011:
recheck

ashu-011:
recheck

ashu-011:
recheck

.zuul.yaml
```diff
@@ -22,7 +22,7 @@
 
 - job:
     name: podified-multinode-edpm-deployment-crc-test-operator
-    parent: podified-multinode-edpm-deployment-crc
+    parent: podified-multinode-hci-deployment-crc-1comp-backends
```
lpiwowar, line 25: @ashu-011 I'm just curious. Why can't we use the edpm job?

lpiwowar (approved):
/approve

It seems to be working. I'm leaving the `/lgtm` up to @kstrenkova.

.zuul.yaml
ashu-011: @lpiwowar We can use but I plan to add some more tests later and the problem is current edpm job doesn't setup a cinder backend due to which all the cinder volume related tests won't work so I switched it with hci job which already setup cinder backend with ceph

ashu-011:
recheck

.zuul.yaml
ashu-011: As discussed with @lpiwowar , Renaming part will be done in a separate PR

ashu-011:
@kstrenkova pls look into this when you get some time :)

kstrenkova (commented):
The patch seems to be working :smile: I checked the horizon-tests and all checks out. The only thing I don't understand is the pre-deploy, but perhaps it has something to do with the need for hci job.

PR #253, opened 2024-11-25
Add ResourceLimits for test pods

lpiwowar:
The test pods spawned by the test operator prior to this change were
executed without any resource limits. This ultimately meant that the
pods could consume an unlimited amount of resources until the limit for
a worker node was reached.

This commit introduces default resource limits for each test pod
spawned by the test operator (tempest, tobiko, ansibletest, horizontest
pod). The default value can be overridden using .Spec.Resources
parameter which is introduced as part of this commit as well [1]:

```
spec:
  resources:
    requests:
      memory: 1Gi
      cpu:    250m
    limits:
      memory: 2Gi
      cpu:    500M
```

[1] https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2617


api/v1beta1/horizontest_types.go
```diff
@@ -28,6 +29,11 @@ import (
 type HorizonTestSpec struct {
 	CommonOptions `json:",inline"`
 
+	// The desired amount of resources that should be assigned to each test pod
+	// spawned using the HorizonTest CR. https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements
+	// +kubebuilder:default:={limits: {cpu: "2000m", memory: "4Gi"}, requests: {cpu: "1000m", memory: "2Gi"}}
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
lpiwowar, line 35: @ashu-011, can you please :pray: suggest the correct Resource Limits for the horizon test pods? It will be possible to override the default value in the HorizonTest CR.

api/v1beta1/tobiko_types.go
```diff
@@ -28,6 +29,11 @@ import (
 type TobikoSpec struct {
 	CommonOptions `json:",inline"`
 
+	// The desired amount of resources that should be assigned to each test pod
+	// spawned using the Tobiko CR. https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements
+	// +kubebuilder:default:={limits: {cpu: "4000m", memory: "4Gi"}, requests: {cpu: "2000m", memory: "2Gi"}}
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
lpiwowar, line 35: @eduolivares , can you please 🙏 suggest the correct Resource Limits for the tobiko test pods? It will be possible to override the default value in the Tobiko CR. For example, in Tempest the amount of resources which is needed is dependent on the number of workers that are used. Maybe it is the same with Tobiko as well?

api/v1beta1/ansibletest_types.go
```diff
@@ -29,6 +30,11 @@ type AnsibleTestSpec struct {
 	CommonOptions         `json:",inline"`
 	CommonOpenstackConfig `json:",inline"`
 
+	// The desired amount of resources that should be assigned to each test pod
+	// spawned using the AnsibleTest CR. https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements
+	// +kubebuilder:default:={limits: {cpu: "4000m", memory: "4Gi"}, requests: {cpu: "2000m", memory: "2Gi"}}
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
lpiwowar, line 36: @eshulman2 , can you please 🙏 suggest the correct default Resource Limits for the ansible test pods? It will be possible to override the default value in the AnsibleTest CR.

lpiwowar (commented):
@eshulman2, @eduolivares, @ashu-011: I would be really grateful :heart:  for your input here. This PR is about setting sane default resource limits [1] for the test pods. The limits are going to be different for each framework therefore I need an estimation from each of you, if possible:).

[1] https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements

api/v1beta1/tobiko_types.go
lpiwowar: Discussed with Eduardo in PM and we agreed on:
1. Setting the default limits to:
```
kubebuilder:default:={limits: {cpu: "8000m", memory: "8Gi"}, requests: {cpu: "4000m", memory: "4Gi"}}
```

2. And setting of the default value of `NumProcesses` to `4`.

lpiwowar:
recheck

api/v1beta1/ansibletest_types.go
lpiwowar: Response from @eshulman2 (through PM):

_... I would say it can be 2G RAM and 400M more less cpu and should work fine (depending on the amount of hosts and the concurrency)._

api/v1beta1/horizontest_types.go
ashu-011: I have tested it in https://github.com/openstack-k8s-operators/test-operator/pull/271 and it seems to be working fine

lpiwowar: Thank you Ashish! :tada:

kstrenkova (changes requested):
I have tested all 4 CRs and they seem to be working. The default values are corresponding to the ones set in code and when changing resource values in CR, the change is propagated into the pod itself.

There is only one request I have regarding this patch and that is adding examples into all 4 CRs of the new parameter Resources. Like so:
```
# resources:
#   limits:
#     cpu: 8000m
#     memory: 4Gi
#   requests:
#     cpu: 4000m
#     memory: 2Gi
```
It lets users know there is an option to change the default resources and gives them overview what the default values are. In my opinion at least :smile:

lpiwowar:
@kstrenkova that's a good idea. Keeping the docs up to date is definitely helpful! Thanks for the review! 

I will update the documentation.

kstrenkova (approved):
Thank you for adding the examples to CRs and documentation! I think now the patch is complete and working, so lgtm.

PR #255, opened 2024-11-28
Enable cinder backend for zuul job

ashu-011:
This PR enables cinder backend with ceph for test-operator zuul job


ashu-011:
recheck

ashu-011:
recheck

PR #256, opened 2024-11-28
Run test pods with readOnlyRootFileSystem: true

lpiwowar:
Up until now the test pods were being spawned with writable root
file system (`readOnlyRootFilesystem: false`). This is against the
best security practices as setting the `readOnlyRootFileSystem` to
`false` increases the size of the attack surface.

This patch ensures that each pod spawned by the test-operator has
the `readOnlyRootFilesystem` set to `true` by default. It is possible
to run a pod with writable roo file system by setting
`privileged: true`.

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/233

### DIFF
--- a/api/bases/test.openstack.org_ansibletests.yaml
+++ b/api/bases/test.openstack.org_ansibletests.yaml
@@ -146,11 +146,11 @@ spec:
                 default: false
                 description: |-
                   Use with caution! This parameter specifies whether test-operator should spawn test
-                  pods with allowedPrivilegedEscalation: true and the default capabilities on
-                  top of capabilities that are usually needed by the test pods (NET_ADMIN, NET_RAW).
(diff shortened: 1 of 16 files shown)


lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar:
The promotion for the podified-antelope-centos9 did not pass yesterday (tcib image does not contain the fix yet):

- https://softwarefactory-project.io/zuul/t/rdoproject.org/buildset/a7e6aed5aea04e80b3c36d5a39b31db1

**-> We have to wait.**

kopecmartin (approved):
lgtm, thanks Lukas

PR #258, opened 2024-12-02
Change suggested namespace to openstack-operators

lpiwowar:
The test-operator should be running in a namespace separate from its
oeprators (test pods). Ultimatelly this means that the test-operator
should be running in the openstack-oeprators namespace with the other
openstack operators.

This patch changes the suggested namespace value in the Test Operator
CSV from openstack to openstack-operators.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2580


lpiwowar:
recheck

lpiwowar:
recheck

PR #259, opened 2024-12-06
Turn off automatic mounting of service account tokens

lpiwowar:
This patch turns off automatic mounting of service account tokens. When a test pod has an access to a service account token it can modify the state of the kubernetes cluster (based on to role assigned to the account). Therefore, it is a good practice to keep the token unmounted by default (unless necessary).

This behavior can be overriden by setting `privileged: true`.


lpiwowar:
https://github.com/openstack-k8s-operators/test-operator/pull/262

PR #262, opened 2024-12-09
Disable insecure parameters by default

lpiwowar:
This patch sets by default:

  - runAsNonRoot: true
  - automountServiceAccountToken: false

These two parameters shouldn't be enabled unless necessary for any pods in the OCP cluster. The runAsNonRoot parameter ensures that the container process does not run as root nor does it switch to root user after it has been spawned.

And the automountServiceAccountToken makes sure that we are not mounting service account tokens to into the test pods. This would allow them to modify the state of the OCP cluster. As of now this is not required by any of the test pods.

This behavior can be overriden by setting privileged: true.


lpiwowar:
ugly PR

PR #264, opened 2024-12-11
Change scope of test-operator to namespace

lpiwowar:
Test-operator is currently designed to be a cluster scoped operator. This means it can watch and modify resources across all OCP cluster.

This patch changes the operator to namespace scoped operator. By default it is going to watch only:

  - openstack-test-operator namespace: This is a namespace where we recommend to install the test-operator. Prior to the installation we recommend to create an OperatorGroup with targetNamespaces value set to openstack-test-operator and openstack.

  - openstack: This is a namespace where the openstack controll plane is deployed. Test-operator requires an access to this namespace in order to read openstack specific CMs and Secrets (e.g., clouds.yaml).

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -101,5 +101,10 @@ spec:
           requests:
             cpu: 10m
             memory: 128Mi
+        env:
+          - name: WATCH_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.annotations['olm.targetNamespaces']
       serviceAccountName: controller-manager
       terminationGracePeriodSeconds: 10
--- a/config/manifests/bases/test-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/test-operator.clusterserviceversion.yaml
@@ -11,7 +11,7 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
-    operatorframework.io/suggested-namespace: openstack
+    operatorframework.io/suggested-namespace: openstack-test-operator
   name: test-operator.v0.0.0
   namespace: placeholder
 spec:
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -1,8 +1,9 @@
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
   name: manager-role
+  namespace: <namespace>
 rules:
 - apiGroups:
   - ""
--- a/config/rbac/role_binding.yaml
+++ b/config/rbac/role_binding.yaml
@@ -1,5 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
   labels:
     app.kubernetes.io/name: clusterrolebinding
@@ -11,7 +11,7 @@ metadata:
   name: manager-rolebinding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
+  kind: Role
   name: manager-role
 subjects:
 - kind: ServiceAccount
--- a/controllers/ansibletest_controller.go
+++ b/controllers/ansibletest_controller.go
@@ -39,6 +39,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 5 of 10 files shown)


kstrenkova (commented):
I have tested this patch and it seems to be running well. The only changes I requested are in the documentation for this because some steps have small mistakes.

I also didn't find the installplans to delete according to documentation :thinking: Maybe it's just a mistake in my testing, but I wanted to mention it here at least.

stuggi:
why you want to change the scope? we can not do it. this needs further discussion

lpiwowar:
/hold

At the moment it looks like that we can not switch the test-operator to project scoped because:

- OLM might be dropping the namespace scoped operators in the near future.
- There might be an issue when using mutating webhooks with namespace scoped operators.

stuggi:
> /hold
> 
> At the moment it looks like that we can not switch the test-operator to project scoped because:
> 
> * OLM might be dropping the namespace scoped operators in the near future.
> * There might be an issue when using mutating webhooks with namespace scoped operators.

when there will be a new api version, conversion webhooks can do the migration of existing objects to the new version. 
Since the CRDs are global objects, the webhooks should be also global to the cluster to be able to perform the migrate.
In theory you could run the webhooks global and limit the operator to be namespaced, but I have not looked into if
thats something easy we could configured to do.
maybe we could revisit our RBAC to set the roles/rolebindings to only the namespaces where there is a deployment.

stuggi:
> > maybe we could revisit our RBAC to set the roles/rolebindings to only the namespaces where there is a deployment.
> 
> What this PR does is that it allows exactly this. It allows the installation of the test-operator in a separate namespace (`openstack-operators`) and then it will create the `roles` and `rolebindings` in the namespaces specified in the `targetNamespace` value in the `OperatorGroup` (`openstack`).
> 
> If you install the test-operator in the `openstack-operators` namespace it should behave like a cluster scoped operator because of the OperatorGroup which is set there [1]:
> 
> ```
> apiVersion: operators.coreos.com/v1
> kind: OperatorGroup
> metadata:
>   name: openstack
>   namespace: openstack-operators
> spec:
>   upgradeStrategy: Default
> status:
>   namespaces:
>     - ""
> ```
> 

how would the clusterroles be managed in this case? you'd have to manually create them?

> But there is the option of installing it in a separate namespace let's say `openstack-test-operators` which has its own OperatorGroup set like this:
> 
> ```
> apiVersion: operators.coreos.com/v1
> kind: OperatorGroup
> metadata:
>   name: openstack
>   namespace: openstack-test-operator
> spec:
>   upgradeStrategy: Default
> status:
>   namespaces:
>     - openstack
> ```
> 
> @stuggi any thoughts, please? I'm 100 % ok with not merging this :) I just want to be sure that there is a justification for it. So do we agree to keep the test-operator ClusterScoped because of?
> 
> * OLM might be dropping the namespace scoped operators in the near future.
> * There might be an issue when using mutating webhooks with namespace scoped operators.

the functionality of the conversion webhooks is crucial, especially for the other service operators. for the test-operator it might be ok to say that you'd have to delete old CRs to unblock installation of a new operator version which brings a new api version, and afterwards you can use the new api, but we can not do it for the other services and from my pov we must be consistent across all operators.

> 
> [1] https://github.com/openstack-k8s-operators/test-operator/pull/264/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R147

lpiwowar:
> how would the clusterroles be managed in this case? you'd have to manually create them?

No, you would not have to create them manually. OLM would create them.

stuggi:
> > how would the clusterroles be managed in this case? you'd have to manually create them?
> 
> No, you would not have to create them manually. OLM would create them.

but you are changing the type here from ClusterRole -> Role

lpiwowar:
> but you are changing the type here from ClusterRole -> Role

Yes, I know. I find this confusing as well. I have test-operator index build [here](quay.io/lpiwowar0/test-operator-index:xxx) which contains test-operator with this PR. If you install test-operator like this:

```
---
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: test-operator-catalog
  namespace: openstack-operators
spec:
  sourceType: grpc
  image: quay.io/lpiwowar0/test-operator-index:xxx
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: test-operator
  namespace: openstack-operators
spec:
  name: test-operator
  source: test-operator-catalog
  sourceNamespace: openstack-operators
```

Then the OLM will create ClusterRole. 

```
The resolved set of selected namespaces is surfaced in an OperatorGroup’s status.namespaces field. A global OperatorGroup’s status.namespace is of length 1 and contains the empty string, "", which signals a consuming operator that it should watch all namespaces.
```
https://olm.operatorframework.io/docs/concepts/crds/operatorgroup/


In contrast if you install test-operator (with this change) like this:
```
---
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: test-operator-catalog
  namespace: openstack-test-operator
spec:
  sourceType: grpc
  image: quay.io/lpiwowar0/test-operator-index:xxx
---
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: test-operator
  namespace: openstack-test-operator
spec:
  targetNamespaces:
    - openstack
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: test-operator
  namespace: openstack-test-operator
spec:
  name: test-operator
  source: test-operator-catalog
  sourceNamespace: openstack-test-operator

```

then only `Roles` and `RolesBindings` will get created for the `openstack` namespace.

lpiwowar:
As per the previous comments, I'm closing this PR. We won't be switching the test-operator to a namespace-scoped one.

PR #266, opened 2024-12-12
 Move away from Jobs to Pods 

lpiwowar:
The test-operator is using Jobs to spawn test pods even though it does
not use any features of this k8s object. Plus usage of the Jobs requires
creation of ServiceAccount in the target namespaces. In order to be able
to create a new, SA the test-oprator has to have a rights to create new
roles and rolebindings which in our case makes the attack surface
larger.

This patch drops the usage of Jobs and moves to Pods.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2749

PR #267, opened 2024-12-13
Remove unnecessary rights

lpiwowar:
In this PR [1] we dropped the usage of the Jobs by the test-operator.
This allows us to drop the rights for:

 - ServiceAccount managing & creation
 - Role managing & creation
 - RoleBinding managing & creation

These rights were only needed because Jobs were required to be spawned
with an extra ServiceAccount that would have elevated privileges in case
the test pods need to run with privileged SecurityContext.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/266

Depends-On: https://github.com/openstack-k8s-operators/test-operator/pull/266


lpiwowar:
recheck

lpiwowar:
This PR will be rebased and ready for a review once this PR gets merged -> https://github.com/openstack-k8s-operators/test-operator/pull/266

controllers/tempest_controller.go
```diff
@@ -58,14 +58,11 @@ func (r *TempestReconciler) GetLogger(ctx context.Context) logr.Logger {
 // +kubebuilder:rbac:groups=test.openstack.org,resources=tempests/finalizers,verbs=update;patch
```
kstrenkova, line 58: I don't know how to add this comment on line 35, but the `common_rbac` definition should be also deleted for this patch to work.

kstrenkova (changes requested):
Without the requested change, I was not able to run the change properly. Other than that, the patch seems to be working.

lpiwowar (commented):
Thank you:)

PR #270, opened 2024-12-17
[zuul] Enable ansibletest and tobiko with stages

lpiwowar:
With this PR in the test-operator role [1] we introduced the concept of stages. By default only tempest stage is enabled. In order for us to continue to test tobiko and ansibletest CR we have to explicitly enable the testing via the cifmw_test_operator_stages parameter.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2374


lpiwowar:
recheck

kopecmartin (approved):
lgtm

PR #276, opened 2025-02-04
Add additional parameters for HorizonTest

ashu-011:
This PR adds additional paramaters to horizontest CR in order to modify the pytest command and projectname xpath based on upstream and downstream dashboard theme

### DIFF
--- a/api/bases/test.openstack.org_horizontests.yaml
+++ b/api/bases/test.openstack.org_horizontests.yaml
@@ -111,6 +111,11 @@ spec:
                   - subPath
                   type: object
                 type: array
+              extraFlag:
+                description: |-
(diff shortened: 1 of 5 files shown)


ashu-011:
recheck

kstrenkova:
To me, the patch seems to be working.  I tried to run horizon tests and it does not create any errors. The code looks correct from the logic and structure perspective as well. After merging the change in tcib, I think this can be also merged.

lpiwowar (approved):
/approve :+1: 

Currently blocked by the issue of two test-operator installations being present in the same job.

lpiwowar:
recheck

kstrenkova:
Seems like this patch is still failing on building documentation. Please do a rebase with main, so this issue gets fixed.

PR #287, opened 2025-03-05
Update documentation about local testing

kstrenkova:
Now that test-operator is installed with other operators, local testing can't be done without some measures. We do not want our local test-operator and the test-operator installed by default to fight for control, so it is needed to disable the default one. This patch documents how this can be achieved.

### DIFF
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -80,8 +80,8 @@ project.
 
 Running Test Operator Locally Outside the Cluster
 -------------------------------------------------
-This is a **quick and easy way** to experiment with the operator during
-development of a new feature.
+This is an easy way to experiment with the operator during development
+of a new feature.
 
 .. code-block:: bash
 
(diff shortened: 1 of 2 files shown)


docs/source/guide.rst
```diff
@@ -90,6 +90,40 @@ development of a new feature.
 Note that after running the following command, you will need to switch to
 another terminal unless you run it in the background.
 
+Now that test-operator is part of the OpenStack product, running changes
```
lpiwowar, line 93: **suggestion (blocking):**
```suggestion
Now that test-operator is automatically deployed in the podified environment, running changes
```

docs/source/guide.rst
```diff
@@ -90,6 +90,40 @@ development of a new feature.
...
+pod, which runs by default. To prevent errors, disable test-operator
+before testing local changes by following these steps.
+
+1. Scale down the openstack-operator deployment
+   This step is necessary because if you only delete the controller managers,
+   the reconcile loop will redeploy them.
+
+.. code-block:: bash
+
+   oc scale deployment openstack-operator-controller-operator -n openstack-operators --replicas=0
```
lpiwowar, line 104: **praise:** I really like this! :)

lpiwowar (changes requested):
Looks good to me! A really nice update of the documentation! :partying_face: 

Just one small nit:).

docs/source/guide.rst
kstrenkova: Ack :smile:

docs/source/guide.rst
kstrenkova: Thanks!

kstrenkova:
I have made the suggested changes and did a small change to the README.md, so the user can now click on a link to get to the _Run Test Operator Locally_ section directly.

PR #300, opened 2025-05-02
Add documentation for ExtraMounts parameter

kstrenkova:
Recently, a new parameter called ExtraMounts was added into test-operator. To ensure that users know how to use it and what fields are needed in the CR, this patch adds documentation on this topic.


lpiwowar (approved):
/approve
/lgtm

LGTM!:) :+1:

PR #308, opened 2025-05-28
Fix setting StorageClass parameter in main section

kstrenkova:
Currently, it is not possible to set storageClass in the main section, if workflow is also set. The only possibility is to set storageClass individually for each step. This patch makes it possible to set storageClass parameter in the main section for every step.

Related-To: https://issues.redhat.com/browse/OSPRH-17024


ASBishop (approved):
I tested this locally and it ltgm, thanks!

lpiwowar (approved):
Nice one @kstrenkova!

/approve

PR #309, opened 2025-05-29
Fix setting Resources parameter in main section

ASBishop:
Update the WorkflowTempestSpec to not assign a default value to the Resources parameter. This lets the Resources value in the main section propagate into every workflow, which eliminates the need to set a custom value in every workflow.

Closes: [OSPRH-16131](https://issues.redhat.com//browse/OSPRH-16131)
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3037

### DIFF
--- a/api/bases/test.openstack.org_ansibletests.yaml
+++ b/api/bases/test.openstack.org_ansibletests.yaml
@@ -2161,13 +2161,6 @@ spec:
                         of tobiko tests).
                       type: boolean
                     resources:
-                      default:
-                        limits:
-                          cpu: 2000m
-                          memory: 2Gi
-                        requests:
-                          cpu: 1000m
-                          memory: 2Gi
                       description: |-
                         The desired amount of resources that should be assigned to each test pod
                         spawned using the AnsibleTest CR. https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements
--- a/api/bases/test.openstack.org_tempests.yaml
+++ b/api/bases/test.openstack.org_tempests.yaml
@@ -2481,13 +2481,6 @@ spec:
                         of tobiko tests).
                       type: boolean
                     resources:
-                      default:
-                        limits:
-                          cpu: 8000m
-                          memory: 4Gi
-                        requests:
-                          cpu: 4000m
-                          memory: 2Gi
                       description: |-
                         The desired amount of resources that should be assigned to each test pod
                         spawned using the Tempest CR. https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements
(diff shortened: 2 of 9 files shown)


kstrenkova (changes requested):
Nice patch! I would suggest also removing defaults from other resource types that use workflow (Tobiko, AnsibleTest). The same way as it is done now.

lpiwowar:
Hey :wave:, 

I'm leaving this review up @kstrenkova as she seems to be doing a nice work :+1:  when doing the review! I'll come back after she adds `lgtm` to do a final check:). 

I'm just adding my observations:

1. With this PR the resource limit is no longer set for `TobikoTest` pods ([1] vs [2] - look for `Limits`). The limit is not intentionally set for `TempestTest` as of now because of [3]. 

2. It seems that the resource limit is not being set for `HorizonTest` and `AnsibleTest` pods as well in the jobs, even though it should IMO. Please, take a look Katka when you have time :pray:. This is not caused by this PR so it should not block it.

My only concern for this PR would be `no.1` (limit not being set for Tobiko pods). `no.2` can be dealt with later.

[1] Job build from this PR -> [HERE](https://logserver.rdoproject.org/61c/rdoproject.org/61c8dffeabd14f65ad99b42a06d478c2/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/tobiko-tests-tobiko-s00-podified-functional/tobiko-tests-tobiko-s00-podified-functional-describe)
[2] Job build before this PR -> [HERE](https://logserver.rdoproject.org/932/rdoproject.org/93294adef0464abe8d24c52a82f96f47/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/tobiko-tests-tobiko-s00-podified-functional/tobiko-tests-tobiko-s00-podified-functional-describe)
[3] https://github.com/openstack-k8s-operators/ci-framework/blob/b34fa2ff99d81822409d0feb48f4173e7c9be5e3/roles/test_operator/defaults/main.yml#L64

kstrenkova (approved):
Thank you @ASBishop for the update and thank you @lpiwowar for the check!

Regarding the first point, I did a small research and the reason for no `limits` set is the line `resource: {}` that is present in the CR. For the defaults to work the `resource` variable must be omitted from the CR (I tested it locally). I am not sure how to fix that yet, but I can create a different ticket to solve it there. This patch looks good to me as it is.

lpiwowar:
/hold

As discussed with @kstrenkova yesterday, this needs to be on hold until the PR on the side of `ci-framework` gets merged [1]. We do not want to replace one bug with another one. But overall, the change should be ok :).

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/3037

kstrenkova:
recheck

PR #312, opened 2025-06-11
Fixup POD_DEBUG var

lmiccini:
According to the spec the default value for POD_DEBUG should be false but this value is not passed properly to the pod, so we end up using the default entrypoint script value (true).


lpiwowar:
recheck

kstrenkova (commented):
I have no problem with these changes, but if the value is not propagated properly I think we should also fix the root of the problem. This is only a workaround for the real issue, if I understood it correctly.

lmiccini:
https://github.com/openstack-k8s-operators/tcib/pull/289

PR #314, opened 2025-06-20
Fix/update common.go to support Tobiko update flag

belolipa:
A flag to indicate whether to update the Tobiko repo for test execution.
If empty or set to "false", no update is performed.

### DIFF
--- a/api/v1beta1/common.go
+++ b/api/v1beta1/common.go
@@ -74,6 +74,13 @@ type CommonOptions struct {
 	// +kubebuilder:default:=""
 	// A URL of a container image that should be used by the test-operator for tests execution.
 	ContainerImage string `json:"containerImage"`
+        
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:default:=""
+        // A flag to indicate whether to update the Tobiko repo for test execution.
+        // If empty or set to "false", no update is performed.
+        TOBIKOUpdateRepo string `json:"tobikoUpdateRepo,omitempty"`
 
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
--- a/api/v1beta1/common.go
+++ b/api/v1beta1/common.go
@@ -74,13 +74,6 @@ type CommonOptions struct {
 	// +kubebuilder:default:=""
 	// A URL of a container image that should be used by the test-operator for tests execution.
 	ContainerImage string `json:"containerImage"`
-        
-	// +operator-sdk:csv:customresourcedefinitions:type=spec
-        // +kubebuilder:validation:Optional
-        // +kubebuilder:default:=""
-        // A flag to indicate whether to update the Tobiko repo for test execution.
-        // If empty or set to "false", no update is performed.
-        TOBIKOUpdateRepo string `json:"tobikoUpdateRepo,omitempty"`
 
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// BackoffLimit allows to define the maximum number of retried executions (defaults to 0).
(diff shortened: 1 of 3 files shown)


api/v1beta1/common.go
```diff
@@ -74,6 +74,13 @@ type CommonOptions struct {
 	// +kubebuilder:default:=""
 	// A URL of a container image that should be used by the test-operator for tests execution.
 	ContainerImage string `json:"containerImage"`
+        
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:default:=""
+        // A flag to indicate whether to update the Tobiko repo for test execution.
+        // If empty or set to "false", no update is performed.
+        TOBIKOUpdateRepo string `json:"tobikoUpdateRepo,omitempty"`
```
kstrenkova, line 83: I would move this value to Tobiko specific file `api/v1beta1/tobiko_types.go`, as the parameter is Tobiko specific. Then the name can be also changed to something like `UpdateRepo` as all the values in that file are tobiko specific, hence no need to add Tobiko in the name.

One more thing is that if we want to indicate if something is either true or false, it's best to use boolean type. You can get inspired how to use boolean here: https://github.com/openstack-k8s-operators/test-operator/blob/ac3151661f643c33bfa781edd89f14b987126c04/api/v1beta1/tobiko_types.go#L61

Another thing, test-operator has a `workflow` feature which you can read on more in test-operator documentation. When we add a new parameter it must be also added as a `Worflow` parameter, you can get inspired by other parameters that are like this.

Additionally, you will have to think about how to propagate this value into TCIB. You can get inspired here: https://github.com/openstack-k8s-operators/test-operator/blob/ac3151661f643c33bfa781edd89f14b987126c04/controllers/tobiko_controller.go#L374

Nice first PR otherwise and good start :D I think you will have no problem with these fixes.

api/v1beta1/tobiko_types.go
```diff
@@ -60,6 +60,12 @@ type TobikoSpec struct {
 	// Boolean specifying whether tobiko tests create new resources or re-use those previously created
 	PreventCreate bool `json:"preventCreate"`
 
+        // +kubebuilder:validation:Optional
+        // +operator-sdk:csv:customresourcedefinitions:type=spec
+        // +kubebuilder:default:=false
+        // Boolean flag to indicate whether to update the Tobiko repo (git pull) before running tests.
```
kstrenkova, line 66: I think the `(git pull)` in the comment is unnecessary.

api/v1beta1/tobiko_types.go
```diff
@@ -140,6 +146,12 @@ type TobikoWorkflowSpec struct {
 	// Boolean specifying whether tobiko tests create new resources or re-use those previously created
 	PreventCreate *bool `json:"preventCreate,omitempty"`
 
+	// +kubebuilder:validation:Optional
+        // +operator-sdk:csv:customresourcedefinitions:type=spec
+        // +kubebuilder:default:=false
+        // Boolean flag to indicate whether to update the Tobiko repo (git pull) before running tests.
```
kstrenkova, line 152: Same here.

controllers/tobiko_controller.go
```diff
@@ -372,6 +372,7 @@ func (r *TobikoReconciler) PrepareTobikoEnvVars(
 	envVars["TOBIKO_LOGS_DIR_NAME"] = env.SetValue(r.GetPodName(instance, workflowStepNum))
 
 	envVars["TOBIKO_TESTENV"] = env.SetValue(instance.Spec.Testenv)
+	envVars["TOBIKO_UPDATE_REPO"] = env.SetValue(strconv.FormatBool(instance.Spec.UpdateRepo))
```
kstrenkova, line 375: I am not sure if this will work, as `strconv` returns `true` or `false`, not `True` or `False`. There should be a testing done to see if it works properly.

kstrenkova (changes requested):
Nice PR update! The requested changes are just minor details along with testing if the patch really works. One more thing that would be great is to update the name of the PR.

belolipa:
Always updating the Tobiko repository is the preferred behavior. Removing the TOBIKO_UPDATE_REPO variable entirely would eliminate unnecessary complexity. We can address this by updating the tcib as part of https://issues.redhat.com/browse/OSPRH-6408.
PR closed, since the change is no longer needed.

PR #318, opened 2025-07-04
Remove frenzyfriday from ci approvers

frenzyfriday:


lpiwowar:
/lgtm

I'm assuming that because this is self-removal that this was thought through.

PR #319, opened 2025-07-04
Edited openstack-k8s-operators dependency bump branch: main

kstrenkova:
This change also includes using the new volume structure that was added manually.


kstrenkova:
I checked locally if the pre-commit check error is only for `AnsibleTest` resource, but it seems that is not the case. After removing `AnsibleTest` resource from the check, `HorizonTest` got the same error. So it is likely it's throwing the error in alphabetical order of the resources and `AnsibleTest` is just first.

Additionally, as the extraMounts parameter is rather new in test-operator, no upstream or downstream jobs are using it. Therefore I think this change could be merged by ignoring the pre-commit check error or finding a way to get rid of it. The change itself should be correct, though.

lpiwowar:
/approve
/lgtm

Nice work with the research on this one @kstrenkova!

lpiwowar:
/test precommit-check

I want to retrigger tide. I removed pre-commit checks from required checks for now. Let's see whether it gets merged by tide. If not, we can force merge.

stuggi:
> * ci/prow/precommit-check

fyi, to get things landed you could override a check like `/override ci/prow/precommit-check` without the need to disable it in some config

PR #322, opened 2025-07-16
Edited openstack-k8s-operators dependency bump branch: 18.0-fr3

kstrenkova:
This is a combined patch consisting of automatic bump and changing the structure of volumes to be compatible with the bump.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3157


stuggi:
recheck

stuggi:
the zuul error seems to be a general issue ? also failing in a test pr https://github.com/openstack-k8s-operators/test-operator/pull/323

kstrenkova:
check-rdo

stuggi:
> @stuggi I am out of ideas what to try :( I tried adding test-operator into the content-provider job, but seems that is not the problem. Even the difference between `main` and `18.0-fr3` branch doesn't seem important enough to cause this problem. Do you have any ideas?

My guess is that its related to the zuul config for this repo, which I am not familiar with.

PR #331, opened 2025-08-27
[tempest re-run] Fix behavior for workflows

sdatko:
In current case, when there is a global value specified
on top level (i.e. non-workflow level, that should be applied
to all workflows) and no workflow-specific value is provided,
that global value will always be overridden with the default
from workflow-level (i.e. the global value will be ineffective
when workflows are used). This commits adjusts it, so value
on workflow-level is optional and has no default – in such case
the for workflow-level the value from top level will be applied
(unless something is explicitly defined on workflow-level).


kstrenkova (approved):
Thank you for the fix! I tested it locally and it worked as expected, so to me it looks good.

sdatko:
https://github.com/openstack-k8s-operators/test-operator/pull/334 needs to land first...

PR #337, opened 2025-09-11
Fix empty AnsibleTest workload key

kstrenkova:
Currently, if user doesn't specify the WorkloadSSHKeySecretName in AnsibleTest CR type, the pod creation is stuck. The reason is that the test-operator is always trying to mount the workload-ssh-secret Secret, but the default is an empty string.


kstrenkova:
recheck

PR #340, opened 2025-09-11
[ansibletest] Add ansibleGitBranch option

elfiesmelfie:
Allow selecting which branch checkout and run the ansible playbook

This allows the playbook to be be run from different branches, which
adds some flexibility:
* The playbook can live on different branches, allowing for versioning
  of the playbook/tests.
* The playbook can be tested in CI prior to merging

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/329

### DIFF
--- a/api/bases/test.openstack.org_ansibletests.yaml
+++ b/api/bases/test.openstack.org_ansibletests.yaml
@@ -63,6 +63,11 @@ spec:
                 default: ""
                 description: AnsibleExtraVars - string to pass parameters to ansible
                 type: string
+              ansibleGitBranch:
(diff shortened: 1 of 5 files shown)


kstrenkova (changes requested):
Hi! Nice looking patch. Could I ask you to add content you get from running `operator-sdk generate kustomize manifests`? Also, could you please update the commit message to be a bit more descriptive (why it is getting added, for example)? Thanks.

kstrenkova (commented):
After the commit message is updated and the related tcib change is merge, I will approve. The change itself seems to be working correctly.

kstrenkova (approved):

PR #350, opened 2025-10-07
Remove GetDefaultBool function

kstrenkova:
The GetDefaultBool function can be replaced with a simple strconv command. This patch is also a part of the test-operator refactor to make the code look more presentable.


kstrenkova:
I have tested the change locally and it seems to be working as expected. The parameters I chose are `Tempest Debug` and `Tobiko Debug`. I tested scenarios when the variable is `true`, `false`, `yes` and their default. The behavior was the same as with the `GetDefaultBool` function.

PR #420, opened 2026-02-25
Rename workflow number variables for clarity

kstrenkova:
While working on a different refactor I noticed that the naming of workflow number variables is a bit misleading. The workflowStepNum variable is currently used for parallel execution, but it is not really clear from the code as it blends in with the workflowStep vairable. This change should better clarify the actual difference between these workflow number variables.


kstrenkova:
recheck

adrianfusco (approved):

belolipa (approved):

PR #432, opened 2026-04-08
[tobiko] Add new parameter SkipRegexList

kstrenkova:
This change introduces a new SkipRegexList parameter to Tobiko. Currently, tests can be skipped by using --skipregex='' parameter in PytestAdopt. However, when having too many entries it becomes cluttered. This parameter helps with keeping the structure readable and user friendly, while keeping the functionality.

### DIFF
--- a/api/bases/test.openstack.org_tobikoes.yaml
+++ b/api/bases/test.openstack.org_tobikoes.yaml
@@ -1378,6 +1378,14 @@ spec:
                       More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
                     type: object
                 type: object
+              skipRegexList:
+                description: |-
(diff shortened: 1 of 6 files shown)


kstrenkova:
recheck

kstrenkova:
recheck

kstrenkova:
I have tested the change locally and it seems to be skipping the tests correctly. For more than one parameter in the pytestAddopts (e.g., `pytestAddopts: -k nova --skipregex='test'` or `non-empty pytestAddopts` with `skipRegexList`) my change in tcib repository will be needed, but I found out that wasn't working before this patch either. So I think it's safe to merge it now and the double-parameter use case will start working once the tcib change is propagated.

eduolivares (approved):

PR #437, opened 2026-04-21
Remove scaffolding TODOs from v1beta1 webhooks

posikoya:
This PR Cleans up the v1beta1 webhook package comments (AnsibleTest, HorizonTest, Tempest, Tobiko) autogenerated

Type: Code hygiene (no functional change)

Description

> Remove comments in internal/webhook/v1beta1/*_webhook.go  files
> Kept +kubebuilder:webhook: markers  and function / method descriptors unchanged so controller-gen and generated manifests keep intended expected behaviour

No logic changes; run existing unit/functional tests and linters as usual.

### DIFF
--- a/internal/webhook/v1beta1/ansibletest_webhook.go
+++ b/internal/webhook/v1beta1/ansibletest_webhook.go
@@ -36,8 +36,7 @@ var (
 	ErrInvalidAnsibleTestType = errors.New("invalid object type for AnsibleTest webhook")
 )
 
-// nolint:unused
-// log is for logging in this package.
+// ansibletestlog is for logging in this package.
 var ansibletestlog = logf.Log.WithName("ansibletest-resource")
 
 // SetupAnsibleTestWebhookWithManager registers the webhook for AnsibleTest in the manager.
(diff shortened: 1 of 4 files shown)


kstrenkova (commented):
Thank you for the patch! Overall, the cleanup is nice, but I would suggest keeping the line break in the comments. When we look at the code, most comments try to be under certain character length for one line. Maybe we should have a check for that in linters :D 

Additionally, before merging, please make sure there is only 1 commit. Multiple commits are acceptable if they implement different features, so usually we want one commit to not have to squash before merging.

Edit: Also one small nitpick. It's nice to stay under 72 characters in the commit/PR title, so in this case I would remove the words in the bracket :D

kstrenkova:
Thank you for the changes! The last thing I think should be updated is the commit message. Please add the text you have on the PR or something more brief to the commit itself. Other than that, looks good to me :+1:

imatza-rh:
LGTM

kstrenkova (approved):
/lgtm
/approve

PR #445, opened 2026-05-06
requeue logic for pending NAD

posikoya:
The operator currently returns a misleading reconcile error when Pod interfaces are not yet ready with assigned IPs. This change introduces a requeue mechanism to allow time for NetworkAttachmentDefinitions (NAD) to be fully provisioned.

### DIFF
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -720,10 +720,10 @@ func (r *Reconciler) VerifyNetworkAttachments(
 			condition.NetworkAttachmentsReadyErrorMessage,
 			err.Error()))
 
-		return ctrl.Result{}, err
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
 }
 
 // EnsureCloudsConfigMapExists ensures that frameworks like Tobiko and Horizon have password values
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -694,6 +694,7 @@ func (r *Reconciler) VerifyNetworkAttachments(
 	conditions *condition.Conditions,
 	networkAttachmentStatus *map[string][]string,
 ) (ctrl.Result, error) {
+	Log := r.GetLogger(ctx)
 	pod, err := r.GetPodIfExists(ctx, instance, workflowStepIndex)
 	if pod == nil {
 		return ctrl.Result{}, err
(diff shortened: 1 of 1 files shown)


internal/controller/common.go
```diff
@@ -720,10 +720,10 @@ func (r *Reconciler) VerifyNetworkAttachments(
 			condition.NetworkAttachmentsReadyErrorMessage,
 			err.Error()))
 
-		return ctrl.Result{}, err
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
```
kstrenkova, line 748: I do think it is a good idea to re-queue, but I am concerned what will happen if the NAD is actually faulty. It seems it will re-queue forever without printing any error in the logs. How would you prevent this?

internal/controller/common.go
```diff
@@ -720,10 +720,10 @@ func (r *Reconciler) VerifyNetworkAttachments(
...
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
```
kstrenkova: Returning an error here might mean that when the `if networkReady` conditions is true, the whole function will return an error, which is not something we want :D

kstrenkova (changes requested):

internal/controller/common.go
posikoya: @kstrenkova in relation to this, i have extended the else statement to have a maxWaitTime var (this maybe added to the spec in future to make it more configurable) which is currently 5 mins. In this way, the re-queue would only silently re-queue up to the max wait time and then exits with the err captured in return ctrl.Result{}, err otherwise it will just return return ctrl.Result{RequeueAfter: time.Second * 10}, nil  and not logging repeatedly for the expected errs we previously knew. Also changing condition.ErrorReason, to "NetworkAttachmentsWaiting", as an untyped string, I thought this was the most fitting description to place, when the user should want to look at the events in reconcile loop rather than the condition which results in error reason

kstrenkova: Looks good to me. The only thing I was thinking about was the `maxWaitTime` which is set to 5 minutes (if it's too little, too much) and the validity of `time.Since(instance.GetCreationTimestamp().Time)`. I will check that out locally, if it works fine, I have no problem with merging it as is!

posikoya: @kstrenkova I do agree with the max time wait, 5mins seems excessive especially from a testing perspective i can definitely adjust this in future

kstrenkova (approved):
Looks good to me. Thank you for your effort :D

PR #448, opened 2026-05-28
Update PodExists function to return pod

kstrenkova:
Replaced PodExists(bool) with GetPodIfExists(*Pod, error) to enable callers to access the pod object while distinguishing between "not found" and actual errors.


sdatko (approved):
/lgtm /approve ❤️

PR #417, opened 2026-02-18
Add basic kuttl tests

kstrenkova:
This PR adds basic kuttl tests and will serve as a base for adding more complex kuttl tests later.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/1137

PR #454, opened 2026-06-15
Fix kuttl cleanup timeout

Valkyrie00:
## Problem

The `tempest-basic` kuttl test fails with `context deadline exceeded` during the `99-cleanup` step after waiting 600 seconds.

The shared `errors-cleanup.yaml` asserted the absence of `openstack-config` ConfigMap and `openstack-config-secret` Secret. These are **infrastructure resources** deployed by `install_yamls` for the CI environment. They are not
created or managed by the test-operator. Other operators running in the `test-operator-kuttl-tests` namespace (keystone, galera, rabbitmq, memcached) keep these resources alive, so they persist after deletion and the error check never passes.

Additionally, the Pod checks in `errors-cleanup.yaml` used `job-name` as the label selector, but the test-operator labels its pods with `instanceName` (see `internal/controller/common.go`). The old selector never matched any test-operator pods, making the check a no-op.

## Fix Proposal

- Remove `openstack-config` ConfigMap and `openstack-config-secret` Secret   from both `cleanup-all-tests.yaml` (delete step) and `errors-cleanup.yaml` The cleanup now only targets test-operator CRs.
- Fix Pod label selectors from `job-name` to `instanceName` so the absence  check actually matches pods created by the test-operator.

Assisted-By: Cursor
Relevance: #417 #452

### DIFF
--- a/test/kuttl/common/cleanup-all-tests.yaml
+++ b/test/kuttl/common/cleanup-all-tests.yaml
@@ -1,12 +1,6 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 delete:
-- apiVersion: v1
-  kind: ConfigMap
-  name: openstack-config
-- apiVersion: v1
-  kind: Secret
-  name: openstack-config-secret
 - apiVersion: test.openstack.org/v1beta1
   kind: Tempest
   name: tempest-sample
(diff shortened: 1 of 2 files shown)


posikoya (approved):
lgtm +1

evallesp (approved):
/lgtm

PR #455, opened 2026-06-18
Add documentation for operator features and security hardening

posikoya:
Document configurable OpenStack config, network attachments, nodeSelector/tolerations, security hardening, privileged mode  capabilities, image creation timeout, and tempest cleanup in crds.rst.  Add ServiceConfigReady condition diagnostics to guide.rst.

To run docs:
`./docs/build-docs.sh`
`sphinx-autobuild docs/source docs/build/html`

### DIFF
--- a/config/samples/test_v1beta1_tempest.yaml
+++ b/config/samples/test_v1beta1_tempest.yaml
@@ -9,6 +9,8 @@ spec:
   # storageClass: local-storage
   # parallel: false
   # debug: false
+  # networkAttachments: []  # list of NADs to attach extra networks to the test pod
+  #                         # if omitted, the pod uses only the default cluster network
 
   # configOverwrite
   # ---------------
@@ -91,6 +93,8 @@ spec:
     #       RAM: 512
     #       disk: 20
     #       vcpus: 1
+    # Set to true to remove all resources created by Tempest during teardown
+    # cleanup: false
 
     # extraRPMs:
     # ----------
--- a/docs/source/crds.rst
+++ b/docs/source/crds.rst
@@ -144,6 +144,33 @@ CRs that can use the workflow section:
 
 * :ref:`ansibletest-custom-resource`
 
+.. _nodeselector-and-tolerations:
+
+NodeSelector and Tolerations
+============================
+Use :code:`nodeSelector` and :code:`tolerations` to control which nodes the
+test pods are scheduled on. The :code:`nodeSelector` ensures the pod only runs
+on nodes with a matching label. The :code:`tolerations` grant permission to run
+on tainted nodes.
+
+Basic Usage
+-----------
+
+.. code-block:: yaml
+
+   spec:
+     # Targets nodes labeled testoperator=true
+     # (oc get node <node-name> --show-labels)
+     nodeSelector:
+       testoperator: "true"
+     # Tolerates nodes tainted with test=true:NoSchedule
+     # (oc describe node <node-name> | grep Taints)
+     tolerations:
+       - key: "test"
+         operator: "Equal"
+         value: "true"
+         effect: "NoSchedule"
+
 ExtraMounts parameter
 =====================
 To correctly use the :code:`ExtraMounts` parameter, follow these steps:
@@ -159,7 +186,7 @@ to be mounted. The name assigned here is later referenced in the
 mounted in the Pod. Each entry should include the name of a volume
 from the :code:`volumes` field and the target mount path.
 
-Example of using the :code:`ExtraMounts` parameter:
+Example test of using the :code:`ExtraMounts` parameter:
 
 .. code-block:: yaml
 
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -201,7 +201,6 @@ Please make sure that you follow the order of the steps:
    :code:`crd`. In such a case, try to delete the :code:`finalizers:`
    section in the output of the :code:`oc edit tempest/tempest-tests`.
 
-
 .. _executing-tests:
 
 Executing Tests
@@ -248,11 +247,31 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
 
 .. code-block:: bash
 
-    oc logs <name of the pod>
+    oc logs <pod-name>
 
 Read :ref:`getting-logs` section if you want to see logs and artifacts
 produced during the testing.
 
+.. _checking-conditions:
+
+Checking Conditions
+-------------------
+Every test CR exposes standard conditions that track its progress:
+
+.. code-block:: bash
+
+   oc get tempest <cr-name> -n openstack -o jsonpath='{.status.conditions}' | jq
+
+Example output of a condition:
+
+.. code-block:: bash
+
+   {
+     "type": "DeploymentReady",
+     "status": "False",
+     "reason": "Requested",
+     "message": "Deployment is running"
+   }
 
 .. _getting-logs:
 
--- a/docs/source/security.rst
+++ b/docs/source/security.rst
@@ -1,0 +1,46 @@
+.. _test-pod-privileges:
+
+===================
+Test Pod Privileges
+===================
+The test-operator follows security best practices by applying a restricted
+security profile to all test pods by default. This section describes the
+defaults and when you might need to change them.
+
+Default Security Profile
+========================
+
+Every test pod runs with the following restrictions:
+
+* **Seccomp**: A :code:`RuntimeDefault` profile filters system calls at the
+  kernel level.
+
+* **Privilege escalation**: Disabled processes cannot gain more privileges
+  than their parent.
+
+* **Root filesystem**: Read-only containers cannot modify their own binaries
+  or write to unexpected locations.
+
+* **Capabilities**: All capabilities are dropped.
+
+* **SELinux level**: Not set by default. Only needed for privileged
+  workflows that write to shared PVCs.
+
+When to Use Privileged Mode
+===========================
+
+.. important::
+   Leave :code:`privileged: false` unless your tests specifically require it.
+
+Set :code:`privileged: true` on the CR spec when your tests need capabilities
+that the default profile blocks. The two capabilities added in privileged mode
+are:
+
+* :code:`NET_ADMIN` allows modifying network configuration such as routing
+  tables, interfaces, and firewall rules.
+
+* :code:`NET_RAW` allows raw socket access for tools like :code:`ping` and
+  :code:`tcpdump`.
+
+Privileged mode is also required when Tempest installs additional RPMs at
+runtime via :code:`extraRPMs`, since that needs a writable root filesystem.
--- a/config/samples/test_v1beta1_ansibletest.yaml
+++ b/config/samples/test_v1beta1_ansibletest.yaml
@@ -16,7 +16,10 @@ spec:
   storageClass: local-storage
   workloadSSHKeySecretName: open-ssh-keys
   ansiblePlaybookPath: playbooks/my_playbook.yaml
+  # git repository URL to clone into the test pod
   ansibleGitRepo: https://github.com/myansible/project
+  # branch, tag, or commit SHA to check out (default: repo's default branch)
+  # ansibleGitBranch: ""
   # containerImage:
   ansibleInventory: |
     localhost ansible_connection=local ansible_python_interpreter=python3
--- a/config/samples/test_v1beta1_tobiko.yaml
+++ b/config/samples/test_v1beta1_tobiko.yaml
@@ -35,6 +35,8 @@ spec:
   # storageClass: local-storage
   # parallel: false
   # debug: false
+  # networkAttachments: []  # list of NADs to attach extra networks to the test pod
+  #                         # if omitted, the pod uses only the default cluster network
   # privateKey: |
   #   <private-key-value>
   # publicKey: |
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -24,6 +24,7 @@ Running Test Operator
    prerequisites.rst
    guide.rst
    crds.rst
+   security.rst
    FAQ.rst
 
 Alternative Ways of Running Tempest Container


docs/source/crds.rst
```diff
@@ -52,6 +52,77 @@ AnsibleTest Custom Resource
...
+
+   spec:
+     ansibleGitRepo: "https://github.com/myorg/openstack-tests"
+     ansibleGitBranch: "v3.0.0"
+     ansiblePlaybookPath: "playbooks/regression.yaml"
+
+The :code:`ansibleGitBranch` field accepts any valid git ref — branch names,
+tag names, or commit SHAs all work.
+
+.. note::
```
kstrenkova: I don't think this note is necessary, this information is already in the first section of this feature.

docs/source/crds.rst
```diff
@@ -52,6 +52,77 @@ AnsibleTest Custom Resource
...
+   If :code:`ansibleGitBranch` is omitted or set to an empty string, the
+   repository's default branch is used.
+
+.. note::
```
kstrenkova: I do not think this explanation is needed as well. The logic related to the merging of the workflows should be in a more visible place. What you have described is not something specific to the AnsibleTest Git Branch, but works like that for all workflow features.

docs/source/crds.rst
```diff
@@ -52,6 +52,77 @@ AnsibleTest Custom Resource
...
+The pod clones the repository, checks out :code:`release-2.1`, then runs the
+playbook.
+
+Run playbooks from a specific tag:
+
+.. code-block:: yaml
+
+   spec:
+     ansibleGitRepo: "https://github.com/myorg/openstack-tests"
+     ansibleGitBranch: "v3.0.0"
```
kstrenkova: Are we sure this works? Even if it does, I would not include this example, I think the one on line 89 is enough to get the picture :D

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
 
 * :ref:`ansibletest-custom-resource`
 
+.. _configurable-openstack-config:
```
kstrenkova: I don't think this is in the right place. I wouldn't say it's such a big feature to be added under this section. It rather makes a bit cluttered with a feature that will not be used as much. I would put it under `Run Tests via Test Operator` and make it a lot shorter.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+
+.. note::
+   The password injection step only applies to HorizonTest and Tobiko. Tempest
+   and AnsibleTest mount the original ConfigMap and Secret directly.
+
+.. _network-attachments:
+
+Network Attachments
+===================
+OpenStack deployments distribute services across multiple isolated networks.
```
kstrenkova: I would drop this paragraph, it doesn't give any useful information imho. The next paragraph about `The networkAttachments field allows users to specify additional....` is enough in that regard.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+
+The :code:`networkAttachments` field allows users to specify additional
+networks that the test pod should be connected to, using Kubernetes
+:code:`NetworkAttachmentDefinition` (NAD) resources provided by the Multus CNI
+plugin.
+
+.. note::
+   NADs must exist in the same namespace as the test CR.
+
+Supported Test Types
```
kstrenkova: This whole section `Supported Test Types` should be just a sentence to the first paragraph saying, they are only supported in `Tempest` and `Tobiko`.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+
+* **Tobiko** - Supported. It creates long-lived OpenStack resources, runs
+  sustained reachability checks, and deliberately disrupts services to validate
+  recovery.
+
+* **HorizonTest** - Not supported.
+
+* **AnsibleTest** - Not supported.
+
+The IP addresses assigned to each interface are visible in the CR's status:
```
kstrenkova: Is this information necessary for the user?

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+     namespace: openstack
+   spec:
+     networkAttachments:
+       - ctlplane
+     containerImage: quay.io/podified-antelope-centos9/openstack-tempest:current-podified
+
+The test pod will receive two network interfaces: the default pod network
+interface, plus one for :code:`ctlplane`.
+
+Per-Workflow-Step Configuration
```
kstrenkova: Again, I don't think talking about the workflow section is necessary here and should be put in a more visible spot, so that it is clear every feature works like that with workflow.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+         networkAttachments:
+           - ctlplane
+           - external
+       - stepName: secondStep-tests
+
+In this example, the :code:`firstStep-tests` step attaches to both
+:code:`ctlplane` and :code:`external`. The :code:`secondStep-tests` step
+inherits the base spec and attaches only to :code:`ctlplane`.
+
+Omitting the Field
```
kstrenkova: Not needed imho, should be clear from how other features work.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+         - 172.17.0.42
+     conditions:
+       - type: NetworkAttachmentsReady
+         status: "True"
+         message: NetworkAttachments ready
+
+Usage
+-----
+
+Prerequisites:
```
kstrenkova: Where did you find this information? What is Multus CNI? :D

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+
+If :code:`networkAttachments` is not specified, the test pod receives only the
+default Kubernetes pod network. The Multus dependency is not required. The
+:code:`NetworkAttachmentsReady` condition will also not appear in the CR's status.
+
+.. _nodeselector-and-tolerations:
+
+NodeSelector and Tolerations
+============================
+The test-operator creates pods to run OpenStack tests. By default, the
```
kstrenkova: Again, this paragraph is not needed imho. It says a lot of stuff, but it's too general and creates clutter.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+Kubernetes scheduler places these pods on any available node based on resource
+availability. In many OpenStack test environments this is not desirable as
+nodes may be partitioned by role, reserved for specific workloads, or
+protected by taints that reject pods without explicit permission.
+
+:code:`nodeSelector` and :code:`tolerations` give users control over which
```
kstrenkova: The first sentence is excellent, the others feel a bit unnecessary.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+nodes the test-operator schedules its pods on. Both fields are available on
+all four test CRDs (Tempest, Tobiko, AnsibleTest, and HorizonTest). When
+omitted, the scheduler uses its default placement logic.
+
+Behaviour Per Test
```
kstrenkova: I don't think this subsection is needed as well.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+     nodeSelector:
+       testoperator: "true"
+     tolerations:
+       - key: "test"
+         operator: "Equal"
+         value: "true"
+         effect: "NoSchedule"
+     containerImage: quay.io/podified-antelope-centos9/openstack-ansible-tests:current-podified
+
+The toleration grants permission to run on the tainted node. The
```
kstrenkova: This could be up in the first paragraph.

docs/source/crds.rst
```diff
@@ -144,6 +215,417 @@ CRs that can use the workflow section:
...
+:code:`nodeSelector` ensures the pod only runs on nodes with the matching
+label.
+
+.. _security-hardening:
+
+Security Hardening
```
kstrenkova: I think this whole security hardening should get a new page and not clutter this one.

docs/source/guide.rst
```diff
@@ -253,6 +253,214 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
 Read :ref:`getting-logs` section if you want to see logs and artifacts
 produced during the testing.
 
+.. _service-config-ready-condition:
+
+ServiceConfigReady Condition
```
kstrenkova: This section is a good idea. I wouldn't necessarily call it `ServiceConfigReady Condition`, but maybe `Checking Conditions`. Then sum up how a user can check the conditions and what does it mean really briefly.

docs/source/guide.rst
```diff
@@ -253,6 +253,214 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
...
+
+.. code-block:: bash
+
+   oc describe tempest my-test -n openstack
+
+Look for the **Conditions:** section in the output.
+
+.. _tempest-rerun-failed-tests:
+
+Tempest Re-run Failed Tests
```
kstrenkova: This section is in the wrong page imho. I would say to remove it all together or brainstorm where to put it. Again, the content is too long with little key information, I would suggest shortening it.

kstrenkova (changes requested):
Thank you for the work you put into this, this PR is much needed!

Saying that, I have left a lot of comments asking for changes. Overall, it is all about the same thing: please make it more compact. I don't think users want to read through sentences that are giving them no new information. I think the documentation should go straight to the point while having the most needed information :D 

The change tackles the right things, however it could use some polishing to match the old style of the documentation and some sections moved to a different place.

docs/source/crds.rst
posikoya: for now i have removed but I was looking for a better area in the docs to place this section

docs/source/crds.rst
posikoya: Multus CNI allows the pods to enable more interfaces and i am not sure in test operator if we can have multiple like ctlplane and and addtion NAD - but i have removed it just to be sure :)

docs/source/crds.rst
posikoya: i added a new section to reflect this but i think i can come up with a better title for the section heading currently

docs/source/guide.rst
posikoya: i removed it but i would say we can do a sync to see where it would suit better to be placed :)

docs/source/security.rst
```diff
@@ -0,0 +1,91 @@
...
+changing routing tables, configuring network interfaces, setting up traffic
+control rules, and modifying firewall rules. Tempest tests that exercise
+neutron functionality may need this to inspect or manipulate the network stack
+from within the test pod.
+
+:code:`NET_RAW` allows the process to create raw network sockets and craft
+arbitrary packets. This is required by tools like :code:`ping` and
+:code:`tcpdump`, consumed by Tempest network connectivity tests.
+
+Behaviour Per Framework
```
kstrenkova: I wouldn't add this `Behaviour Per Framework` section. First of all, I am not sure if the text is correct for every resource. Then I am also concerned that it is just text clutter.

docs/source/crds.rst
```diff
@@ -144,6 +194,188 @@ CRs that can use the workflow section:
...
+container uses this value when polling Glance for the image status after
+upload. If the image does not reach :code:`active` within the specified number
+of seconds, the script exits with an error and the test pod fails.
+
+.. note::
+   The default value is 300 seconds (5 minutes).
+
+.. _tempest-cleanup:
+
+Tempest Cleanup
```
kstrenkova: I think you can remove the whole Tempest Cleanup section. I don't think it's difficult to understand what the cleanup does, it's also just setting one boolean value.

If we want to let people know of it's existence, we can add it into the `Tempest Custom Resource` as a comment like:

```
# Set to true to remove all resources created by Tempest during teardown
#    cleanup: false
```

docs/source/crds.rst
```diff
@@ -144,6 +194,188 @@ CRs that can use the workflow section:
...
+     nodeSelector:
+       testoperator: "true"
+     tolerations:
+       - key: "test"
+         operator: "Equal"
+         value: "true"
+         effect: "NoSchedule"
+     containerImage: quay.io/podified-antelope-centos9/openstack-ansible-tests:current-podified
+
+.. _image-creation-timeout:
```
kstrenkova: I would remove this section as well. It is only Tempest specific and everyone knows what a timeout means. I think the section overall is too detailed around how it works as well. We don't need to explain every little detail to the reader, the documentation should explain briefly how it works and how to use it.

docs/source/guide.rst
```diff
@@ -253,6 +273,34 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
...
+
+.. code-block:: bash
+
+   oc describe tempest my-test -n openstack
+
+Look for the **Conditions:** section in the output:
+
+.. code-block:: text
+
+   Conditions:
```
kstrenkova: When you run `oc describe` do you see this output? :D You need to talk about something that the user will experience exactly.

I would recommend either talking about using `oc get tempest tempest-tests -o jsonpath='{.status.conditions}' | jq` or talk about what you can really see under **Conditions**, e.g.
```
Conditions:
  Type                        Status
  PodReadyToStartContainers   False 
  Initialized                 True 
  Ready                       False 
  ContainersReady             False 
  PodScheduled                True 
```

docs/source/guide.rst
```diff
@@ -253,6 +273,34 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
...
+     Type                          Status  Reason      Message
+     ----                          ------  ------      -------
+     InputReady                    True    Ready       Input validation passed
+     ServiceConfigReady            True    Ready       Service config ready
+     DeploymentReady               False   Requested   Deployment is running
+     NetworkAttachmentsReady       True    Ready       Network attachments ready
+     Ready                         False   Requested   Deployment is running
+
+Each condition is **True** (passed), **False** (failed — read the Message for
```
kstrenkova: This part could be either removed or simplified (plus put next to the first sentence in this section). I think saying the first thing that failed is the problem should be obvious 😅

docs/source/crds.rst
```diff
@@ -52,6 +52,56 @@ AnsibleTest Custom Resource
    :language: yaml
 
 
+.. _git-branch-selection-for-ansibletest:
+
+Git Branch Selection for AnsibleTest
```
kstrenkova: I would remove this section and add it to the `AnsibleTest Custom Resource` as a comment. I don't think the feature is very difficult to understand and doesn't need much text to be explained.

docs/source/crds.rst
kstrenkova: You moved it to the other file, but the remaining text is still here :D

docs/source/crds.rst
```diff
@@ -144,6 +194,188 @@ CRs that can use the workflow section:
...
+to point :code:`openStackConfigMap` to their original ConfigMap without
+worrying about the injected password.
+
+.. note::
+   The password injection step only applies to HorizonTest and Tobiko. Tempest
+   and AnsibleTest mount the original ConfigMap and Secret directly.
+
+.. _network-attachments:
+
+Network Attachments
```
kstrenkova: Network attachments are another feature that I would just comment on in `Tempest Custom Resource` and `Tobiko Custom Resource`. It is not difficult to understand, just adding another network, which can be explained in a simple comment.

docs/source/security.rst
```diff
@@ -0,0 +1,91 @@
+.. _security-hardening:
+
+==================
+Security Hardening
```
kstrenkova: I would perhaps name this section `Test pod privileges` or something similar. I think talking about how it is set to follow security best practices is nice, just not piling too much (not useful) information on the user.

Overall this page is a good idea, but some parts are too specific like:
```
Two unnecessary host mounts (:code:`/etc/localtime` and
:code:`/etc/machine-id`) were also removed — .....
```
It should talk about things that actually affect the reader. Explain when to use it, when not.

kstrenkova (changes requested):
Thank you for your changes :D The structure is better now, but I left more comments to further reduce a bit of clutter. I hope my comments make sense, but if you are unsure, please let me know!

config/samples/test_v1beta1_tempest.yaml
```diff
@@ -91,6 +93,8 @@ spec:
     #       RAM: 512
     #       disk: 20
     #       vcpus: 1
+    # Set to true to remove all resources created by Tempest during teardown
+    # cleanup: true
```
kstrenkova: Could you please put `cleanup: false`, as it has default value false? The examples use their defaults for consistency :D Besides stuff like extraMounts which is more of an example.

docs/source/guide.rst
```diff
@@ -253,6 +273,15 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
...
 
+.. _checking-conditions:
+
+Checking Conditions
+-------------------
+Every test CR exposes standard conditions that track its progress:
+
+.. code-block:: bash
+
+   oc get tempest tempest-tests -n openstack -o jsonpath='{.status.conditions}' | jq
```
kstrenkova: The text before at line 271 uses `<name of the pod>`, so maybe that would be better to use for consistency. Or if there is `tempest-tests` used elsewhere, it can be kept like this.

Maybe it would be nice to show one example of how the output looks, but not all the conditions, so it is not crowded 🤔 What do you think?

docs/source/crds.rst
```diff
@@ -144,6 +144,36 @@ CRs that can use the workflow section:
...
+     namespace: openstack
+   spec:
+     nodeSelector:
+       testoperator: "true"
+     tolerations:
+       - key: "test"
+         operator: "Equal"
+         value: "true"
+         effect: "NoSchedule"
+     containerImage: quay.io/podified-antelope-centos9/openstack-ansible-tests:current-podified
```
kstrenkova: Could you please remove the `containerImage` line here? We do not want explicit URLs of this in the documentation, as it might be changed in the future + it is not needed in the example for it to work.

docs/source/security.rst
```diff
@@ -0,0 +1,49 @@
...
+  than their parent.
+
+* **Root filesystem**: Read-only — containers cannot modify their own binaries
+  or write to unexpected locations.
+
+* **Capabilities**: All capabilities are dropped.
+
+* **SELinux level**: No default — users set this explicitly when needed.
+
+No configuration is required to get these defaults. They apply automatically
```
kstrenkova: I would remove this sentence, it reads as "no need to set defaults, as they are the defaults".

docs/source/security.rst
```diff
@@ -0,0 +1,49 @@
...
+to Tempest, Tobiko, HorizonTest, and AnsibleTest pods.
+
+When to Use Privileged Mode
+===========================
+Set :code:`privileged: true` on the CR spec when your tests need capabilities
+that the default profile blocks. The two capabilities added in privileged mode
+are:
+
+* :code:`NET_ADMIN` — allows modifying network configuration (routing tables,
```
kstrenkova: This is just a my own opinion, but I don't like the hyphens in such cases. Is it possible to rewrite it a bit without it?

Do we know for a fact the info about `Tempest neutron plugin tests and Tobiko fault injection scenarios.`? I would remove it if we are not sure, also for the `Needed by Tempest connectivity tests and Tobiko reachability checks.` Just saying it is enabled should be enough :D

docs/source/security.rst
```diff
@@ -0,0 +1,49 @@
...
+  Tobiko fault injection scenarios.
+
+* :code:`NET_RAW` — allows raw socket access for tools like :code:`ping` and
+  :code:`tcpdump`. Needed by Tempest connectivity tests and Tobiko
+  reachability checks.
+
+Privileged mode is also required when Tempest installs additional RPMs at
+runtime via :code:`extraRPMs`, since that needs a writable root filesystem.
+
+If your tests do not exercise network infrastructure or install extra
```
kstrenkova: I would remove the `(the default)` and make this part start with something marking it as an important information. Maybe even move it to the top of the `When to Use Privileged Mode` because it should be clear that it should stay `false` unless necessary.

docs/source/guide.rst
```diff
@@ -201,6 +201,26 @@ Please make sure that you follow the order of the steps:
    :code:`crd`. In such a case, try to delete the :code:`finalizers:`
    section in the output of the :code:`oc edit tempest/tempest-tests`.
 
+.. _configurable-openstack-config:
+
+Configurable OpenStack Config
+------------------------------
+Test pods authenticate against OpenStack using two resources mounted into
```
kstrenkova: I am really pondering whether to add this section. I can't really double-check the information without researching it like, in which cases would user want to set it otherwise, will they always have `clouds.yaml` and `secure.yaml`, what does it mean that  "operator automatically injects the password" if the user decides that they want to use their custom files.

Overall, if you don't mind I would drop this section and we can discuss how to add it later (there will be more updates to the documentation anyway). Or you can research when it is used, what purpose would name changing bring, etc. I will leave it up to you 👍

docs/source/crds.rst
kstrenkova: Could you update it so that the first sentence is not saying something so general? So it is more straight to the point.

Additionally, I see `toleration` is not in `:code:`, but `nodeSelector` is and there is a space missing between `availability` and `The`. There is also `the tainted nod` which should be `node` I assume and also the use of `the` is a bit weird in the sentence 🤔 Is there a specific tainted node?

docs/source/crds.rst
```diff
@@ -144,6 +144,36 @@ CRs that can use the workflow section:
...
+
+   apiVersion: test.openstack.org/v1beta1
+   kind: AnsibleTest
+   metadata:
+     name: ansible-validate
+     namespace: openstack
+   spec:
+     nodeSelector:
+       testoperator: "true"
+     tolerations:
```
kstrenkova, line 168: Could you test this example locally if it works? I never set tolerations and nodeSelector, so I am not sure if that is accurate :D

kstrenkova (changes requested):
Thank you for the PR! The overall content is a lot better, you applied the changes I asked for correctly 😊 For my review now, I focused more on specific parts of the text where I think small improvements are needed. So I think the update should be hopefully simple this time around. Thanks for your patience, I think we are really close to reaching an agreement!

docs/source/guide.rst
posikoya: done - i added an example of one using a json block instead

docs/source/security.rst
posikoya: done - i have removed the hypens. For the scenarios, i do not know this for a fact - so for now i have removed it as i went off the pr which introduced security concepts, but i can add a task to test this functionality for sure but for now i have removed it as we can add it back later as we continue to build out the docs :)

docs/source/guide.rst
posikoya: done - I have removed it for now in the interest of having a base docs from these changes as that we are both happy with. i think also 100% that yes when we build out the docs more it could be added it again and we can discuss on this but i would need to do more research here and then test

docs/source/crds.rst
posikoya: i still need to conduct the test here - i will work on this asap :D

docs/source/security.rst
kstrenkova: Thank you! I wouldn't add specific information like that for now, but it could be beneficial for a documentation update later as you said.

kstrenkova (approved):
Lgtm. The change is really well put together and helps to cover more test-operator features in the documentation.

docs/source/guide.rst
```diff
@@ -248,11 +247,31 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
...
 
+.. _checking-conditions:
+
+Checking Conditions
+-------------------
+Every test CR exposes standard conditions that track its progress:
+
+.. code-block:: bash
+
+   oc get tempest <pod-name> -n openstack -o jsonpath='{.status.conditions}' | jq
```
belolipa: The `oc get tempest` command operates on the Custom Resource, not the pod. Using `<pod-name>` here is inconsistent with the earlier section and will confuse users.

posikoya: Hey Oleg - I get what you mean ,  <pod-name>  was used because the conditions reflect the state of the test pod that gets spawned during the test, However, since oc get tempest queries the CR object, not the pod directly, the placeholder should reference whatever is in the CR's metadata. in this case shaill we say tempest-name or cr-name or just name in this case

posikoya: i have update it now @belolipa  is it okay to resolve this comment?

belolipa: Please change:
pod-name → tempest-cr-name or cr-name

docs/source/guide.rst
```diff
@@ -248,11 +247,31 @@ You should see a pod with a name like :code:`tempest-tests-xxxxx`.
...
 
+.. _checking-conditions:
+
+Checking Conditions
+-------------------
+Every test CR exposes standard conditions that track its progress:
+
+.. code-block:: bash
+
+   oc get tempest <cr-name> -n openstack -o jsonpath='{.status.conditions}' | jq
```
posikoya, line 263: @belolipa done - it has been changed here

belolipa (approved):

PR #456, opened 2026-06-18
add envtests for extraMounts feature

posikoya:
The extraMounts feature allows mounting additional volumes (ConfigMaps, Secrets, PVCs) into test pods via propagation filtering.

Add extraMounts envtests for all four controllers (AnsibleTest, Tempest, Tobiko, HorizonTest) that verify the reconciler correctly propagates extra volumes and volumeMounts to the pod spec.i

### DIFF
--- a/test/functional/ansibletest_controller_test.go
+++ b/test/functional/ansibletest_controller_test.go
@@ -116,4 +116,112 @@ var _ = Describe("AnsibleTest controller", func() {
 			Expect(pod.Name).ToNot(BeEmpty())
 		})
 	})
+
+	Context("extraMounts", func() {
+		BeforeEach(func() {
+			openstackConfigMap, openstackSecret := CreateCommonOpenstackResources(namespace)
+			Expect(k8sClient.Create(ctx, openstackConfigMap)).Should(Succeed())
+			Expect(k8sClient.Create(ctx, openstackSecret)).Should(Succeed())
+
+			testOperatorConfigMap := CreateTestOperatorConfigMap(namespace)
+			Expect(k8sClient.Create(ctx, testOperatorConfigMap)).Should(Succeed())
+		})
+
+		When("AnsibleTest is created with extraMounts", func() {
+			BeforeEach(func() {
+				CreateExtraConfigMap(namespace, ExtraConfigMapName)
+
+				spec := GetDefaultAnsibleTestSpec()
+				spec["extraMounts"] = BuildExtraMountsSpec("AnsibleTest",
+					GetDefaultConfigMapExtraMount())
+
+				DeferCleanup(th.DeleteInstance, CreateAnsibleTest(ansibleTestName, spec))
+			})
+
+			It("should add extra volume and volumeMount to the pod", func() {
+				pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+				ExpectPodHasConfigMapVolume(pod, ExtraConfigVolName, ExtraConfigMapName)
+				ExpectPodHasVolumeMount(pod, ExtraConfigVolName, ExtraConfigMountPath)
+			})
+		})
+
+		When("AnsibleTest is created with Secret as the source of extraMount", func() {
+			BeforeEach(func() {
+				CreateExtraSecret(namespace, ExtraSecretName)
+
+				spec := GetDefaultAnsibleTestSpec()
+				spec["extraMounts"] = BuildExtraMountsSpec("AnsibleTest",
+					GetDefaultSecretExtraMount())
+
+				DeferCleanup(th.DeleteInstance, CreateAnsibleTest(ansibleTestName, spec))
+			})
+
+			It("should add secret based extra volume and volumeMount to the pod", func() {
+				pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+				ExpectPodHasSecretVolume(pod, ExtraSecretVolName, ExtraSecretName)
+				ExpectPodHasVolumeMount(pod, ExtraSecretVolName, ExtraSecretMountPath)
+			})
+		})
+
+		When("AnsibleTest is created with multiple extraMounts configmap and secret", func() {
+			BeforeEach(func() {
+				CreateExtraConfigMap(namespace, ExtraConfigMapName)
+				CreateExtraSecret(namespace, ExtraSecretName)
+
+				spec := GetDefaultAnsibleTestSpec()
+				spec["extraMounts"] = BuildExtraMountsSpec("AnsibleTest",
+					GetDefaultConfigMapExtraMount(),
+					GetDefaultSecretExtraMount(),
+				)
+
+				DeferCleanup(th.DeleteInstance, CreateAnsibleTest(ansibleTestName, spec))
+			})
+
+			It("should add all extra volumes and volumeMounts to the pod", func() {
+				pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+				ExpectPodHasConfigMapVolume(pod, ExtraConfigVolName, ExtraConfigMapName)
+				ExpectPodHasSecretVolume(pod, ExtraSecretVolName, ExtraSecretName)
+				ExpectPodHasVolumeMount(pod, ExtraConfigVolName, ExtraConfigMountPath)
+				ExpectPodHasVolumeMount(pod, ExtraSecretVolName, ExtraSecretMountPath)
+			})
+		})
+
+		When("AnsibleTest is created with no propagation field", func() {
+			BeforeEach(func() {
+				CreateExtraConfigMap(namespace, ExtraConfigMapName)
+
+				spec := GetDefaultAnsibleTestSpec()
+				spec["extraMounts"] = BuildExtraMountsSpec("",
+					GetDefaultConfigMapExtraMount())
+
+				DeferCleanup(th.DeleteInstance, CreateAnsibleTest(ansibleTestName, spec))
+			})
+
+			It("should add extra volume and volumeMount when propagation is omitted", func() {
+				pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+				ExpectPodHasConfigMapVolume(pod, ExtraConfigVolName, ExtraConfigMapName)
+				ExpectPodHasVolumeMount(pod, ExtraConfigVolName, ExtraConfigMountPath)
+			})
+		})
+
+		When("AnsibleTest created with extraMounts is using the wrong propagation type", func() {
+			BeforeEach(func() {
+				CreateExtraConfigMap(namespace, ExtraConfigMapName)
+
+				spec := GetDefaultAnsibleTestSpec()
+				spec["extraMounts"] = BuildExtraMountsSpec("Tempest",
+					GetDefaultConfigMapExtraMount())
+
+				DeferCleanup(th.DeleteInstance, CreateAnsibleTest(ansibleTestName, spec))
+			})
+
+			It("should not add extra volume and volumeMount to the pod", func() {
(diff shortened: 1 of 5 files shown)


test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +117,385 @@ var _ = Describe("AnsibleTest controller", func() {
...
+	When("AnsibleTest is created with extraMounts", func() {
+		BeforeEach(func() {
+			openstackConfigMap, openstackSecret := CreateCommonOpenstackResources(namespace)
+			Expect(k8sClient.Create(ctx, openstackConfigMap)).Should(Succeed())
+			Expect(k8sClient.Create(ctx, openstackSecret)).Should(Succeed())
+
+			testOperatorConfigMap := CreateTestOperatorConfigMap(namespace)
+			Expect(k8sClient.Create(ctx, testOperatorConfigMap)).Should(Succeed())
+
+			extraConfigMap := &corev1.ConfigMap{
```
kstrenkova: Could you create a helper function that would create this `extraConfigMap`? Something like `CreateExtraMountConfigMap`?

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +117,385 @@ var _ = Describe("AnsibleTest controller", func() {
...
+					Namespace: namespace,
+				},
+				Data: map[string]string{
+					"config.conf": "key=value",
+				},
+			}
+			Expect(k8sClient.Create(ctx, extraConfigMap)).Should(Succeed())
+
+			spec := GetDefaultAnsibleTestSpec()
+			spec["extraMounts"] = []map[string]any{
```
kstrenkova: Same here. It could be a helper function that would have `AnsibleTest` as a variable so it could be used for other resource types as well :D

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +117,385 @@ var _ = Describe("AnsibleTest controller", func() {
...
+	When("AnsibleTest is created with Secret as the source of extraMount", func() {
+		BeforeEach(func() {
+			openstackConfigMap, openstackSecret := CreateCommonOpenstackResources(namespace)
+			Expect(k8sClient.Create(ctx, openstackConfigMap)).Should(Succeed())
+			Expect(k8sClient.Create(ctx, openstackSecret)).Should(Succeed())
+
+			testOperatorConfigMap := CreateTestOperatorConfigMap(namespace)
+			Expect(k8sClient.Create(ctx, testOperatorConfigMap)).Should(Succeed())
+
+			extraSecret := &corev1.Secret{
```
kstrenkova: Same here as with the config map, but maybe it could be `CreateExtraMountSecret`?

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +117,385 @@ var _ = Describe("AnsibleTest controller", func() {
...
+					},
+				},
+			}
+			DeferCleanup(th.DeleteInstance, CreateAnsibleTest(ansibleTestName, spec))
+		})
+
+		It("should add extra volume and volumeMount to the pod", func() {
+			pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+
+			foundVolume := false
```
kstrenkova: This could be a helper function as well, something that looks for volumes and is just called with the volume name as parameter (plus all other parameters needed).

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +117,385 @@ var _ = Describe("AnsibleTest controller", func() {
...
+					Namespace: namespace,
+				},
+				StringData: map[string]string{
+					"secret.conf": "key=value",
+				},
+			}
+			Expect(k8sClient.Create(ctx, extraSecret)).Should(Succeed())
+
+			spec := GetDefaultAnsibleTestSpec()
+			spec["extraMounts"] = []map[string]any{
```
kstrenkova: Same here, but with secret.

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +117,385 @@ var _ = Describe("AnsibleTest controller", func() {
...
+					Namespace: namespace,
+				},
+				Data: map[string]string{
+					"data.conf": "key=value",
+				},
+			}
+			Expect(k8sClient.Create(ctx, extraDataMap)).Should(Succeed())
+
+			spec := GetDefaultAnsibleTestSpec()
+			spec["extraMounts"] = []map[string]any{
```
kstrenkova: This one also needs a helper function, the code it too long to just be here and make clutter :D

kstrenkova (changes requested):
Thank you for the PR! I have tried running it locally and it is working as expected :D 

Overall the functionality and scale of the tests is nice, however I would like you to introduce a few helper functions into `test/functional/base_test.go` to reduce the repetition and code clutter. I have commented on a few, but please note that all of the code creating resources and searching for a name match should be helper functions, even when I have not marked them.

Note: I think Claude will have a quick solution for this (moving code into helper functions) if you feel like it's too repetitive to fix it all.
Note2: The name of the commit should start with a verb to keep good practice.

test/functional/ansibletest_controller_test.go
Valkyrie00: +1 and I would suggest creating this function to the `base_test` file.

test/functional/ansibletest_controller_test.go
Valkyrie00: +1, e.g.
```
func BuildConfigMapExtraMountSpec(propagation string, volName string, cmName string, mountPath string) []map[string]any { ... }
func BuildSecretExtraMountSpec(propagation string, volName string, secretName string, mountPath string) []map[string]any { ... }
```

test/functional/ansibletest_controller_test.go
Valkyrie00: +1

Valkyrie00:
Hey @posikoya, the test scenarios themselves are well-chosen and functionally correct.
The blocking issue is the extensive code duplication. I agree with all of @kstrenkova's comments, especially regarding the helper functions, extracting those would make the PR dramatically smaller and test code more maintainable 👍

test/functional/ansibletest_controller_test.go
posikoya: ack - I had added helper generic funcs to the base_test file and i will update the rest of the tests. the current changes are currently passing on ansible test and has cut down alot of the duplication here. also ramping up on this test framework, i added some additions such as context [https://medium.com/@dees3g/testing-with-ginkgo-and-gomega-1f1ecc8407a8](url) to reduce duplicated calls also. really appreciate the reviews here both :)

posikoya: done

test/functional/ansibletest_controller_test.go
posikoya: done

test/functional/ansibletest_controller_test.go
posikoya: done

test/functional/ansibletest_controller_test.go
posikoya: done

test/functional/ansibletest_controller_test.go
posikoya: I agree also i moved the logic to the helper funcs file in base_test and done

test/functional/ansibletest_controller_test.go
posikoya: Done: I added the func `BuildExtraMountsSpec` with return signature `return []map[string]any{
		{"extraVol": []map[string]any{extraVol}}, 
	}` to account for the scenario of multiple mounts (mount 0 - mount N-1) with different types (configmap + secret)

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +116,126 @@ var _ = Describe("AnsibleTest controller", func() {
...
+			})
+
+			It("should add extra volume and volumeMount when propagation is omitted", func() {
+				pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+				ExpectPodHasConfigMapVolume(pod, "extra-config-vol", "extra-config")
+				ExpectPodHasVolumeMount(pod, "extra-config-vol", "/etc/extra-config")
+			})
+		})
+
+		When("AnsibleTest created with extraMounts using the wrong propagation type", func() {
```
kstrenkova: I would add `is` here (`is using the wrong...`) and in all the other "wrong propagation" tests :D

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +116,126 @@ var _ = Describe("AnsibleTest controller", func() {
...
+			Expect(k8sClient.Create(ctx, testOperatorConfigMap)).Should(Succeed())
+		})
+
+		When("AnsibleTest is created with extraMounts", func() {
+			BeforeEach(func() {
+				CreateExtraConfigMap(namespace, "extra-config")
+
+				spec := GetDefaultAnsibleTestSpec()
+				spec["extraMounts"] = BuildExtraMountsSpec("AnsibleTest",
+					ExtraMount{VolName: "extra-config-vol",
```
kstrenkova: These values for ExtraMount are same across the files, so could you maybe create constants from them. Something like this in `test/functional/base_test.go`?

```
const (
      ExtraConfigMapName = "extra-config"
      ExtraConfigVolName = "extra-config-vol"
      ExtraConfigMountPath = "/etc/extra-config"
      ExtraSecretName = "extra-secret"
      ExtraSecretVolName = "extra-secret-vol"
      ExtraSecretMountPath = "/etc/extra-secret"
  )
```

kstrenkova: Once you have the constants, I would move the ExtraMount creation to `test/functional/base_test.go` as well. So you can call it like:

```
spec["extraMounts"] = BuildExtraMountsSpec("AnsibleTest",
                GetDefaultConfigMapExtraMount())
```

test/functional/ansibletest_controller_test.go
```diff
@@ -116,4 +116,126 @@ var _ = Describe("AnsibleTest controller", func() {
...
+			It("should add secret based extra volume and volumeMount to the pod", func() {
+				pod := GetTestOperatorPod(namespace, ansibleTestName.Name)
+				ExpectPodHasSecretVolume(pod, "extra-secret-vol", "extra-secret")
+				ExpectPodHasVolumeMount(pod, "extra-secret-vol", "/etc/extra-secret")
+			})
+		})
+
+		When("AnsibleTest is created with multiple extraMounts configmap and secret", func() {
+			BeforeEach(func() {
+				CreateExtraConfigMap(namespace, "extra-config")
```
kstrenkova: Use the new constants here and in the `It("should add all extra volumes and volumeMounts to the pod",` as well.

kstrenkova (changes requested):
The changes are nice overall. I have a few more comments to reduce the duplication a bit more, but those should be a quick fix. Once they are addressed, I am okay with merging this change.

test/functional/ansibletest_controller_test.go
posikoya: done

test/functional/ansibletest_controller_test.go
posikoya: done

test/functional/ansibletest_controller_test.go
posikoya: done

kstrenkova (commented):
/lgtm
/approve

kstrenkova (approved):

PR #458, opened 2026-06-22
Set ImagePullPolicy to IfNotPresent on test pods

abays:
BuildTestPod creates raw Pods directly without going through lib-common workload helpers, so the recently added pod.SetPullPolicyDefaults does not apply. When ImagePullPolicy is not explicitly set and the image tag is "latest" or unset, Kubernetes defaults the policy to Always. Explicitly set PullIfNotPresent to avoid unnecessary image pulls.


stuggi (approved):
/lgtm

PR #463, opened 2026-07-01
Distinguish pending from running pods

kstrenkova:
This is a first step to adding timeouts for pending pods. Previously, it wasn't possible to distinguish a pod that was pending from a pod that is running. Now it is possible to distinguish them.

This lets us handle pending pods differently and later add timeout logic to cancel tests early if pods are stuck pending.

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>


posikoya (approved):
lgtm

PR #471, opened 2026-07-21
Remove hardcoded passwords following install_yamls#1158

kstrenkova:
A recent change [1] in install_yamls repository replaced hardcoded passwords with dynamically generated secrets. To keep test-operator up to date, it is needed to also update/remove hardcoded passwords, which is the goal of this PR. The new dynamic passwords also break EnsureCloudsConfigMapExists which injected "12345678" into clouds.yaml when empty, but that value no longer matches the real admin password.

This patch removes the workaround and has Tobiko and HorizonTest mount the original openstack-config ConfigMap directly. For HorizonTest, the password is read from osp-secret thanks to a ci-framework patch [2]. For Tobiko, a new way of supporting the password from secure.yaml is being worked on [3].

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/1158
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/4067
[3] https://review.opendev.org/c/x/tobiko/+/998010

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/4083

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/openstack-k8s-operators/lib-common/modules/storage v0.6.1-0.20260717092345-ab1ee7b97c67
 	github.com/openstack-k8s-operators/test-operator/api v0.0.0-00010101000000-000000000000
 	go.uber.org/zap v1.28.0
-	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.33.13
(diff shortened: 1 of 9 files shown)


Valkyrie00 (approved):

posikoya (approved):

imatza-rh (approved):
/lgtm
/approve

PR #474, opened 2026-07-23
Remove unused fields from HorizonTest CR

kstrenkova:
The ProjectName, User, Password fields were never read by the controller. In TCIB it is hardcoded as "horizontest" regardless of what was set in the CR.

Normally, I would recommend deprecation for parameters, but since these parameters never worked from the start, I suggest a removal. There is also no need for them to be parameters, as the Horizon team agreed they do not have to be overwritten and can stay hardcoded.

### DIFF
--- a/api/bases/test.openstack.org_horizontests.yaml
+++ b/api/bases/test.openstack.org_horizontests.yaml
@@ -1252,34 +1252,15 @@ spec:
                   - extraVol
                   type: object
                 type: array
-              flavorName:
-                default: m1.tiny
-                description: FlavorName is the name of the OpenStack flavor to create
-                  for Horizon tests.
-                type: string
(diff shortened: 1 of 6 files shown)


posikoya (commented):
@kstrenkova this is a great MR and very much needed! :) while going through the initial review everything lgtm. There is only a small thing I noticed, that there may remain the CSV entries in `config/manifests/bases/test-operator.clusterserviceversion.yaml` for the 7 removed fields (around lines 305-373) for` flavorName, horizonTestDir, imageUrl, logsDirectoryName, password, projectName, user`

posikoya (approved):
/approve
/lgtm

adrianfusco (approved):

PR #483, opened 2026-07-31
Switch build-test-operator workflow to shared reusable-build-operator

stuggi:
Delegate image/bundle/index building to openstack-k8s-operators-ci/reusable-build-operator.yaml, matching the pattern used by other operators. Drops the now-unused create_bundle.sh/create_opm_index.sh scripts since the reusable workflow handles that logic internally.


rebtoor (approved):
/approve
/lgtm

